PR #10, opened 2021-11-25
Implement DeviceCopy for Complex numbers

sebcrozet:
This PR adds an optional implementation of `DeviceCopy` for Complex numbers. This can be very valuable for scientific computing. The `num_complex::Complex` type is already `Copy` and `#[repr(C)]`.

### DIFF
--- a/crates/cust/Cargo.toml
+++ b/crates/cust/Cargo.toml
@@ -12,6 +12,7 @@ cust_raw = { path = "../cust_raw", version = "0.11.2"}
 bitflags = "1.2"
 cust_derive = { path = "../cust_derive", version = "0.1" }
 vek = { version = "0.15.1", optional = true }
+num-complex = { version = "0.4", optional = true }
 
 [build-dependencies]
 find_cuda_helper = { path = "../find_cuda_helper", version = "0.1" }
--- a/crates/cust/src/memory/mod.rs
+++ b/crates/cust/src/memory/mod.rs
@@ -286,3 +286,6 @@ impl_device_copy_vek! {
     CubicBezier2, CubicBezier3,
     Quaternion,
 }
+
+#[cfg(feature = "num-complex")]
+unsafe impl<T: DeviceCopy> DeviceCopy for num_complex::Complex<T> {}


RDambrosio016:
I'm kind of wondering if this is the wrong approach, it might be better to instead PR a CUDA feature to num to make it derive DeviceCopy. Just so we don't end up with hundreds of features.

crates/cust/Cargo.toml
```diff
@@ -12,6 +12,7 @@ cust_raw = { path = "../cust_raw", version = "0.11.2"}
 bitflags = "1.2"
 cust_derive = { path = "../cust_derive", version = "0.1" }
 vek = { version = "0.15.1", optional = true }
+num-complex = { version = "0.4", optional = true }
```
cuviper, line 15: You probably want `default-features = false`.

RDambrosio016:
The plan currently is to actually move DeviceCopy out of cust and into cust_core which will build on gpu targets, that way you don't need tons of cfg's for structs. So i think i will hold off on merging this until then, that way we can decide if we want this in cust_core or in num-complex. Im personally for the latter considering how many math crates could be useful for the GPU, but i am not sure.

sebcrozet:
@RDambrosio016 I think that @cuviper [makes a very good point](https://github.com/rust-num/num-complex/issues/97#issuecomment-980442207): since `Rust-CUDA` will likely evolve a lot in the near future, it is best to have the `cust` (or `cust_core`) depend on `num-complex` instead of the other way round. Then we could revisit this decision in a few years once `Rust-CUDA` becomes more stable.

The core issue with having `num-complex` depend on `cust` (or `cust_core`) is that each minor version bump (which we can expect to happen fairly frequently in the near future) of `cust` would imply a minor version bump of `num-complex`, which in turns implies a minor version bump of the whole ecosystem relying directly or indirectly on `num-complex` (even if they don’t care about CUDA).
Having `cust` depend on `num-complex` on the other hand would have little consequences on the ecosystem versioning considering that `num-complex` is already fairly stable it this point (with only 3 minor version bumps during the past two years).

RDambrosio016:
Hmm thats a good point, tho we will need to move this to cust_core later

PR #31, opened 2021-12-24
Fixes to Dockerfile

CircArgs:
Apologies for the minor PRs in succession. I realized there were some minor issues with the Dockerfile that are now fixed and running the add example successfully seems to confirm it is working properly.

Details: The primary issue was that building succeeded, but running failed to create a context. This was somehow related to `ld` not finding the shared objects despite being in `LD_LIBRARY_PATH`. I reverted to `ldconfig` and this fixed the issue. 

Starting this PR as a draft to hear opinions on whether the Dockerfile should optionally include optix with a build argument. I have not investigated this much but I can look into it.

### DIFF
--- a/guide/src/guide/getting_started.md
+++ b/guide/src/guide/getting_started.md
@@ -5,7 +5,7 @@ This section covers how to get started writing GPU crates with `cuda_std` and `c
 ## Required Libraries
 
 Before you can use the project to write GPU crates, you will need a couple of prerequisites:
-- [The CUDA SDK](https://developer.nvidia.com/cuda-downloads), version `11.2` or higher. This is only for building
+- [The CUDA SDK](https://developer.nvidia.com/cuda-downloads), version `11.2` or higher (and the appropriate driver - [see cuda release notes](https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html)) . This is only for building
 GPU crates, to execute built PTX you only need CUDA 9+.
 
 - LLVM 7.x (7.0 to 7.4), The codegen searches multiple places for LLVM:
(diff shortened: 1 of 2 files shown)


RDambrosio016:
The optix SDK needs an nvidia dev account so probably not, the headers can be downloaded from a link just fine but im pretty sure the sdk installer does other things too

guide/src/guide/getting_started.md
```diff
@@ -90,7 +92,7 @@ Now we can finally start writing an actual GPU kernel.
 <details>
   <summary>Expand this section if you are not familiar with how GPU-side CUDA works</summary>
 
-Firstly, we must explain a couple of things about GPU kernels, specifically, how they are executed. GPU Kernels (functions) are the entry point for executing anything on the GPU, they are the functions which will be executed from the CPU. GPU kernels do not return anything, they write their data to buffers passed into them.
+Firstly, we must explain a couple of things about GPU kernels, specifically, how they are executed. GPU Kernels (functions) are the entry oint for executing anything on the GPU, they are the functions which will be executed from the CPU. GPU kernels do not return anything, they write their data to buffers passed into them.
```
RDambrosio016, line 95: 🤔

CircArgs: whoops must have deleted that somehow. Thanks for pointing it out. Should be fixed now

CircArgs:
> The optix SDK needs an nvidia dev account so probably not, the headers can be downloaded from a link just fine but im pretty sure the sdk installer does other things too

I did notice it required a dev account. I wasn't sure if the install script they provide could just be bundled with this and be used to build in optix. That's cool though I think this Dockerfile is complete for now then

PR #39, opened 2022-01-11
docs: add `output` file hack to tips

Nilstrieb:
When I built my project, `rustc_codegen_nvvm` rebuilt on every subsequent build of my program, even if I just changed the CPU crate code. Someone on the discord server shared their trick on how to fix it, by touching the `output` file in the `target` or `rustc_codegen_nvvm` and it worked for me, so I added it to the guide.


RDambrosio016:
Hmmm i think i would rather try and fix this build.rs issue than document a hack for fixing it

Nilstrieb:
I would add it temporarily, fix the issue somehow, and then remove it again. It's certainly good to know about it right now

PR #42, opened 2022-02-07
Remove unnecessary "An" in README.md

yerke:
Remove unnecessary "An" in README.md


RDambrosio016:
The "An" here is grammatically correct, although it should not be capitalized

PR #44, opened 2022-02-14
Make DeviceSlice a newtype around [()]

beepster4096:
This fixes some unsoundness and fixes some of DeviceSlice's methods and docs not making any sense (such as from_raw_parts_mut)


crates/cust/src/memory/device/device_slice.rs
```diff
@@ -221,14 +236,15 @@ impl<T: DeviceCopy + Pod> DeviceSlice<T> {
     /// In total it will set `sizeof<T> * len` values of `value` contiguously.
     #[cfg_attr(docsrs, doc(cfg(feature = "bytemuck")))]
     pub fn set_8(&mut self, value: u8) -> CudaResult<()> {
-        if self.ptr.is_null() {
-            return Ok(());
-        }
-
```
RDambrosio016, line 227: Why were these checks removed? They are needed for ZST device slices, since zero-width allocation in cuda is not possible (and makes literally no sense)

crates/cust/src/memory/device/device_slice.rs
```diff
@@ -426,8 +439,16 @@ pub trait DeviceSliceIndex<T: DeviceCopy> {
...
+    unsafe fn get_unchecked(self, slice: &DeviceSlice<T>) -> &DeviceSlice<T>;
+    fn index(self, slice: &DeviceSlice<T>) -> &DeviceSlice<T>;
+
+    /// Indexes into this slice without checking if it is in-bounds.
+    ///
+    /// # Safety
+    ///
+    /// The range must be in-bounds of the slice.
+    unsafe fn get_unchecked_mut(self, slice: &mut DeviceSlice<T>) -> &mut DeviceSlice<T>;
+    fn index_mut(self, slice: &mut DeviceSlice<T>) -> &mut DeviceSlice<T>;
```
RDambrosio016, line 451: The reason for reimplementing `SliceIndex` was that it was impossible to make a `&DeviceSlice<T>` because itd be returning a temporary value, reimplementing deviceslice as a DST should fix this and make `DeviceSliceIndex` useless, at that point we can implement it using normal `SliceIndex` (tho not sure how that behaves with a ZST slice). Or keeping `DeviceSliceIndex` but implementing `Index` and `IndexMut` again

RDambrosio016 (changes requested):
Great PR, couple of things that need to be sorted out, and the changes need to be documented in the changelog

crates/cust/src/memory/device/device_slice.rs
DrMeepster: These checks only checked if the pointer itself was null, which is not actually possible with references. I'll replace them with length checks then.

crates/cust/src/memory/device/device_slice.rs
DrMeepster: normal `SliceIndex` will be UB with `[T]` and won't change the pointer position correctly with `[()]` so we still need `DeviceSliceIndex`, but I'll impl `Index(Mut)`

RDambrosio016: I see, i think we should just remove the index method we have already and reimplement the index traits. Not *too* excited about re-breaking the deviceslice API lmao but this will be a 0.4 thing.

crates/cust/src/memory/device/device_slice.rs
RDambrosio016: Hmmm not sure if that will work because zst deviceslices still must track their length correctly

RDambrosio016: Honestly i think we should just straight up remove support for ZST device allocations, they make no sense, there is no reason why you would want to pass a ZST to the GPU in the form of a device slice. In that case we can insert an assert on allocation that ensures the type is not a ZST, either that or return an InvalidAllocation error

RDambrosio016 (commented):
Looks good, just needs changelog entries

DrMeepster:
Alright, changelog is upodated and merge conflicts are resolved

LegNeato:

Is this still relevant?

LegNeato:
Thanks for this PR from a long time ago @beepster4096 ! 🍻

PR #47, opened 2022-02-16
Add memcpy_dtoh in cust

### LINKED ISSUE #45: Expose cuMemcpyDtoH_v2 in cust
In the latest version (`0.3.0`) of cust `cuMemcpyHtoD_v2` was exposed as `cust::memory::memcpy_htod`. This avoids the need for reaching into `cust_raw` and allows using `cust::error::CudaResult` instead of error codes.

Should `cuMemcpyDtoH_v2` be exposed in a similar manner to enable the features above for device to host memcpy as well?

kjetilkjeka:
As discussed in and fixes #45 

Success case is tested locally on my computer.

### DIFF
--- a/crates/cust/CHANGELOG.md
+++ b/crates/cust/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 Notable changes to this project will be documented in this file.
 
+## Unreleased
+
+ - Add `memory::memcpy_dtoh` to allow copying from device to host.
+
 ## 0.3.2 - 2/16/22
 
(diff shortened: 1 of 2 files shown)


RDambrosio016:
Thanks for the PR 🙂, it just needs to be documented in Changelog.md in `crates/cust`, add another `## Unreleased` entry and add this change

kjetilkjeka:
Squashed the update of the changelog into my previous commit. 

I tried following your previously used format, let me know if I should change it in any ways.

PR #62, opened 2022-03-16
Feat: Add multi-head attention backward passes and convolution fused bias-act

frjnn:


RDambrosio016:
ah yes, i know what that means

PR #64, opened 2022-03-17
Add pitched malloc and 2d memcpy to cust

### LINKED ISSUE #56: How should the future cust::memory look like? (Originally: Should pitched alloc and 2D memcpy be exposed in cust?)
This issue was really going to be about malloc pitch and 2D, but then I realized maybe the whole `cust::memory` needed a little refurbish. 

Cuda gives better performance when rows are aligned in a specific manner. For 2D arrays (e.g. images) it is common to use `cudaMallocPitch` for allocation and `cudaMemcpy2D` for copying. In addition there exists corresponding functions for 3D which makes things a bit more complex as the cuda array type must be used. 

The first question is: Should these be exposed, in some way, in cust? I think the answer is yes, or else there would be unnecessarily difficult to write code that is as performant as when using for host code C/C++.

The second question is: How should these functions be exposed in cust? Cuda is all over the place when it comes to naming, argument lists, return values and even how things really work. Should we follow the cuda runtime API to make it as similar as possible for the C++ crowd? Or should try to make `cust::memory` as coherent as possible? 

I understand both sides, but it seems to me the goal for this project is to make Cuda in Rust as good as it can be without being afraid of diverging from how it works in C++. The logical solution might then be to try to improve on naming and function signatures. 

When I started looking at this I realized that the current `cust::memory` module is not very unified either. The alloc and memcpy functions operate of different types (cust::DevicePointer<T> vs CUdeviceptr), size specifications (bytes vs elements) and constraints (when allocing T must be `DeviceCopy`). How should the feature complete `cust::memory` look like?

kjetilkjeka:
Fixes #56 

Add pitched malloc and 2d memcpy to `cust::memory`. 

The memcpy_2d functions is working on elements in width/height instead of bytes. This does not correspond to what the other `memcpy` functions are currently doing but corresponds to what `malloc` (including `malloc_pitched`) is currently doing.

I see many good arguments for using bytes and many good arguments for using elements. The important thing is that the interface is consistent. My assumption is that it is more important with consistency between malloc and memcpy than between memcpys. Making the module completely consistent is outside the scope of this PR as discussed in #56 

I also updated cust from 2018 edition to 2021 edition. This was mainly to get `TryInto` automatically into scope. I hope that was Ok.

### DIFF
--- a/crates/cust/CHANGELOG.md
+++ b/crates/cust/CHANGELOG.md
@@ -5,6 +5,7 @@ Notable changes to this project will be documented in this file.
 ## Unreleased
 
  - Add `memory::memcpy_dtoh` to allow copying from device to host.
+ - Add support in `memory` for pitched malloc and 2D memcpy between device and host.
 
 ## 0.3.2 - 2/16/22
 
--- a/crates/cust/Cargo.toml
+++ b/crates/cust/Cargo.toml
@@ -6,7 +6,7 @@ authors = [
     "Riccardo D'Ambrosio <rdambrosio016@gmail.com>",
     "Brook Heisler <brookheisler@gmail.com>",
 ]
-edition = "2018"
+edition = "2021"
 license = "MIT OR Apache-2.0"
 description = "High level bindings to the CUDA Driver API"
 repository = "https://github.com/Rust-GPU/Rust-CUDA"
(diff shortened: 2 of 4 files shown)


therishidesai:
Is there any path forward on getting this merged in?

LegNeato:

Is this still relevant?

LegNeato:
As these are just additions, we'll merge.

PR #69, opened 2022-05-05
Fixed `impl_glam|mint|vek` feature flags being an illusion of choice

WilliamVenner:
I noticed that it is not possible to disable these features because `cust_core` (which `cust` depends on) uses them as default features but exposes no way to turn them off.


WilliamVenner:
I believe so!

LegNeato:
Looks reasonable. I highly doubt anyone is relying directly on `cust_core`.

Thanks again for the PR and the prompt response! 🍻

PR #71, opened 2022-05-11
Add support for getting the UUID of a device

vmx:
`Device::uuid()` was added to get the UUID of a device via the underlying
`cuDevuceGetUuid()` call.

### DIFF
--- a/crates/cust/src/device.rs
+++ b/crates/cust/src/device.rs
@@ -346,7 +346,7 @@ impl Device {
         unsafe {
             cuDeviceGetUuid(&mut cu_uuid, self.device).to_result()?;
         }
-        let uuid: [u8; 16] = cu_uuid.bytes.map(|byte| byte.to_ne_bytes()[0]);
+        let uuid: [u8; 16] = cu_uuid.bytes.map(|byte| byte as u8);
         Ok(uuid)
     }
 
--- a/crates/cust/src/device.rs
+++ b/crates/cust/src/device.rs
@@ -327,6 +327,29 @@ impl Device {
         }
     }
 
+    /// Returns the UUID of this device.
+    ///
+    /// # Example
+    /// ```
+    /// # use cust::*;
+    /// # use std::error::Error;
+    /// # fn main() -> Result<(), Box<dyn Error>> {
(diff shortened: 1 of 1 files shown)


crates/cust/src/device.rs
```diff
@@ -327,6 +327,29 @@ impl Device {
...
+    /// let device = Device::get_device(0)?;
+    /// println!("Device UUID: {:?}", device.uuid()?);
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn uuid(self) -> CudaResult<[u8; 16]> {
+        unsafe {
+            let mut cu_uuid = CUuuid { bytes: [0i8; 16] };
+            cuDeviceGetUuid(&mut cu_uuid, self.device).to_result()?;
+            let uuid: [u8; 16] = ::std::mem::transmute(cu_uuid.bytes);
```
RDambrosio016, line 348: Using a transmute for just `i8` -> `u8` seems a bit odd, just using `array.map()` would be better.

vmx: Fixed with the most recent push (I also fixed the commit message to match the usual style).

crates/cust/src/device.rs
```diff
@@ -327,6 +327,29 @@ impl Device {
...
+    /// println!("Device UUID: {:?}", device.uuid()?);
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn uuid(self) -> CudaResult<[u8; 16]> {
+        let mut cu_uuid = CUuuid { bytes: [0i8; 16] };
+        unsafe {
+            cuDeviceGetUuid(&mut cu_uuid, self.device).to_result()?;
+        }
+        let uuid: [u8; 16] = cu_uuid.bytes.map(|byte| byte.to_ne_bytes()[0]);
```
RDambrosio016, line 349:
```suggestion
        let uuid: [u8; 16] = cu_uuid.bytes.map(|byte| byte as u8);
```

vmx:
I can also squash and rebase the commit if you prefer that.

RDambrosio016:
I usually squash commits anyways, so this is fine :)

Issue #19, opened 2021-12-06
build failure ->  rustc_codegen_nvvm(build) 


1617176084:
When I add this package
`cuda_builder = { version = "0.2", path = "./Rust-CUDA/crates/cuda_builder" }
`
And compiled, and then they started a bunch of bugs, I had LLVM_ and CUDA installed

 rustc_codegen_nvvm(build)
![image](https://user-images.githubusercontent.com/7008899/144815551-1e14173e-3a2c-4d0c-a14d-c1b405dc56b0.png)

RDambrosio016:
I may have forgotten to mention it in the guide but you must copy the rust-toolchain file to your own project root, that is because the codegen uses rustc internals so it cannot be stable until it is upstreamed in rustc

PR #76, opened 2022-06-30
Fix guide book to mention `rust-toolchain` and small build fix

jac-cbi:
Using the provided Dockerfile, my new GPU project failed because I didn't know about these two needed changes from the guidebook.  Add them to the guidebook so that no one else gets caught by it.


jac-cbi:
This should fix issue [19](https://github.com/Rust-GPU/Rust-CUDA/issues/19)

RDambrosio016 (approved):

PR #87, opened 2022-09-16
Add bridging PTX code for cooperative_groups API

thedodd:
Overall, this system will work roughly as follows:
- Wrap a `cc::Builder` to compile the bridging code on demand as part of a `sys` crate. This will use nvcc under the hood and we will pass along any other needed flags.
- Folks that need the bridging code will then use cust::link::Linker to link the bridging PTX code with their PTX.
- We can wrap this up in a nice builder interface, maybe even make it part of the current `cuda_builder` crate as a nice optional feature + method.

Linking the PTX on this branch along with a custom kernel of mine which is using the new `cuda_std::cg` module, is all working nicely. However, it currently deadlocks invocations because we have not exposed the cooperative launch interface. **This is exactly what I would expect, and is what the cuda docs say will happen.** Exposing the `cuLaunchCooperativeKernel` function should be pretty simple, which is an outstanding task for this PR.

**An important note:** the generated PTX included in this branch includes the same core ASM which is seen when building an all C++ demo app with nvcc (when output as PTX).

LOTS TO DO STILL!
- [ ] Update the new cuda_std_cg crate to use `cc` to compile the C++ bridging code on demand, and produce the PTX path as output. Maybe wrap the path and the bytes in a nice intermediate object in case folks want to use it in different ways.
- [ ] MAJOR: update cust to expose the `cuLaunchCooperativeKernel` in a nice interface. We can add the cooperative multi device bits later, along with all of the other bits from the cooperative API.
- [ ] Remove the Justfile. I was only using it for POC testing.
- [ ] Remove the PTX code from this branch, currently it is just here for reference.


thedodd:
@RDambrosio016 whenever you get some time (no rush), let me know what you think. I am testing this out as I go on a fairly large project of mine which brought about this need in the first place.

Overall, the bridging code is quite simple. I've given an outline of how I think this should be exposed overall. Let me know what you think, happy to modify things as I go.

Also, for this first pass, I would like to keep focused only on the grid-level components of the cooperative groups API, as well as the basic cooperative launch host-side function. We can add multi-device and the other cooperative group components later.

RDambrosio016:
This looks neat, but if im not mistaken, those functions map to single PTX intrinsics directly, wouldn't it be easier to use inline assembly? though i haven't actually looked into this so im not sure if they map to more than one PTX instruction

thedodd:
> wouldn't it be easier to use inline assembly?

I started down that path at first, and for a few of the pertinent functions the corresponding PTX was clear. I was using a base C++ program compiled down to PTX to verify in addition to cross-referencing with the [PTX ISA spec](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html). However, I will say, many of the interfaces were not as clear, and this seemed to be a potentially more reliable way to generated the needed code.

Perhaps we can replace some of the clear interfaces with some ASM instead. Happy to iterate on this in the future.

LegNeato:

Is this still relevant?

PR #90, opened 2022-09-29
Fix: make UUID retrieval more portable

vmx:
The `CUuuid` byte type is `std::os::raw::c_char` and it depends on the platform, whether it is `i8` or `u8` in Rust. With just passing in `0` without a type identifier, both cases are handled. This makes the code more portable.

### DIFF
--- a/crates/cust/src/device.rs
+++ b/crates/cust/src/device.rs
@@ -342,7 +342,7 @@ impl Device {
     /// # }
     /// ```
     pub fn uuid(self) -> CudaResult<[u8; 16]> {
-        let mut cu_uuid = CUuuid { bytes: [0i8; 16] };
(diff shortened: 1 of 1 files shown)


vmx:
A friendly ping if this PR could perhaps get a review (and ideally a release :) This would enable me to get off a fork and use the upstream version.

LegNeato:

Is this still relevant?

LegNeato:
This looks reasonable so I am going to merge. I am still ramping on the code, why not use `c_char` directly? Thanks again!

vmx:
> I am still ramping on the code, why not use `c_char` directly?

It depends on how you view a UUID. I'd say it's a list of bytes. While `c_char` is practically a byte, one could also say it's a character which would be a semantic difference.

The other thing is that I think it's a good idea to use Rust types in the public API where possible, rather then C specific types.

PR #93, opened 2022-10-25
On NixOS find cudatoolkit via `which nvcc`, fixes #92.

### LINKED ISSUE #92: Support NixOS / add path fallback logic via `which nvcc`
Right now the crate `find_cuda_helper()` has a few hard-coded paths it checks on Linux (e.g., `/opt/cuda` or `/usr/local/cuda`). 

On NixOS however the CUDA toolkit gets installed to a location like `/nix/store/n3mnxpif0zxs4ws1pw8spj68l0gzcr9z-cudatoolkit-11.7.0`, including `libs` and `includes`, and all `bin` commands are made available in the current PATH. 

So to compile CUDA crates on NixOS there should be a fallback in `find_cuda_lib_dirs()` that, for example, searches for the presence of `nvcc` (e.g., by running `which nvcc`) and uses its parent as the return value of that method. (Note, not saying this is the best solution for NixOS, please chime in if you know something better).

ralfbiedert:


LegNeato:

Is this still relevant?

ralfbiedert:
I haven't touched this in 2 years, but I'd assume this is still an issue on NixOS.

LegNeato:
This seems reasonable to me. I don't use nix, so let's YOLO. Thanks again for the PR from a long time ago 😆

PR #101, opened 2023-01-29
LLVM Prebuilt Binary Download Fix

giulianbiolo:
On windows I've been experiencing this problem where I can't seem to download the tar.gz without returning an error, this fixed it for me.


RDambrosio016:
Disabling SSL verification is really dangerous (docs say this too), so i'd much rather not, i think you should verify what problems you are having with SSL and fix them on your side.

PR #105, opened 2023-03-20
Fix: cast caused double free (use mem::forget)

DoeringChristian:
Hi,

i found a double free in the cast function of the DeviceBuffer class.
The following code would not work:
```rust
        let x = vec![0f32; 100].as_slice().as_dbuf().unwrap();
        dbg!(&x);
        println!("{:#x?}", x.as_device_ptr().as_raw());
        // dbg!(x.as_host_vec());
        let x = x.cast::<u8>();
        dbg!(&x);
        let x = x.cast::<f32>();
        dbg!(&x);
        dbg!(x.as_host_vec());
```

This returns with CUDA_ERROR_INVALID_VALUE.
Looking at the trace using Nvidia Nsight Compute it can be seen that there occurs a double free.

Screenshot from 2023-03-20 09-12-56
It turns out that when the cast function returns it drops self and frees the buffer.
The fix is to use `std::mem::forget` on self.
```rust
            let new_len = (size_of::<A>() * self.len) / size_of::<B>();
            let ret = Ok(DeviceBuffer {
                buf: self.buf.cast(),
                len: new_len,
            });
            unsafe{std::mem::forget(self);} // this is necessary to prevent double free
            ret
```
However I don't know if there is a better way than using `std::mem::forget`.

### DIFF
--- a/crates/cust/src/memory/device/device_buffer.rs
+++ b/crates/cust/src/memory/device/device_buffer.rs
@@ -314,7 +314,7 @@ impl<A: DeviceCopy + Pod> DeviceBuffer<A> {
     /// whole number of elements. Such as `3` x [`u16`] -> `1.5` x [`u32`].
     /// - If either type is a ZST (but not both).
     #[cfg_attr(docsrs, doc(cfg(feature = "bytemuck")))]
-    pub fn try_cast<B: Pod + DeviceCopy>(self) -> Result<DeviceBuffer<B>, PodCastError> {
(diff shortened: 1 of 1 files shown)


LegNeato:
Hello @DoeringChristian ! We are [rebooting this project](https://rust-gpu.github.io/blog/2025/01/27/rust-cuda-reboot/). Sorry for your PR not getting merged!

Is this still relevant?

DoeringChristian:
Hey, cool to hear! I haven't checked if it is in a long time. You can close it if you want to.

LegNeato:
I think we should take this, as it look localized and was debugged and fixed in the past.  Thanks again!

PR #128, opened 2025-01-08
add new cuda default path for 24.04

zeroecco:
in Ubuntu 24.04 cuda libs are installed in directory:
```
/usr/lib/cuda
```
I propose that this be added to default "search" locations.


LegNeato:
CI failure is unrelated.

PR #137, opened 2025-02-02
Take an &Event in Stream::wait_event

### LINKED ISSUE #110: wait_event method in Stream
I have encountered an issue related to the wait_event method in the Stream struct. It appears that when the method is called, it consumes the Event object, resulting in the event being dropped each time.

Considering that generating events is not cheap, I would like to understand the rationale behind passing ownership of the event and dropping it on each wait. Is there a specific reason for not passing an event reference instead, allowing the event to be reused?

Thanks!

juntyr:
Fixes #110


juntyr:
@LegNeato Thoughts?

LegNeato:
This looks reasonable to me.

PR #138, opened 2025-02-02
Allow Stream::add_callback to run on context error again

juntyr:
See https://stackoverflow.com/a/58173486


juntyr:
@LegNeato What are your thoughts? Alternatively, we could also re-expose the old method with a new name and keep the existing one.

LegNeato:
Nope, this works! Thanks again, sorry for the slow response..I am traveling and have spotty availability

PR #146, opened 2025-02-27
Updated crates to use new versions of Dependencies

Schmiedium:
Most of the project builds and uses the newest version of all the dependencies. The only exception is the nvvm codegen crate which still doesn't build

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,4 +1,7 @@
 [workspace]
+
+resolver = "2"
+
 members = [
   "crates/*",
   "crates/optix/examples/ex*",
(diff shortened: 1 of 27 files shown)


juntyr (approved):
LGTM! As a minor nit, could you use std::cell::OnceCell now that our MSRV is high enough?

Schmiedium:
Yes! Good catch, I'll update that tomorrow

Schmiedium:
So it looks like rustc_nvvm_codegen wasn't using once_cell anyway? I removed the dependency from cargo.toml, so it should be good to go

PR #143, opened 2025-02-18
Cargo build bug fix.

Jerryy959:
This pr for running cargo build bug-fix.
This pull request includes changes to the `crates/rustc_codegen_nvvm/build.rs` file and the `rust-toolchain` file to re-enable support for a specific target and update the Rust toolchain channel. The most important changes are as follows:

Re-enabling target support:

* [`crates/rustc_codegen_nvvm/build.rs`](diffhunk://#diff-67ee338edb5803a94f2d3e53ed6d3c3783e2bc4059eaafa6a343a42c8f880548L57-R57): Re-enabled the `x86_64-unknown-linux-gnu` target by uncommenting the corresponding line in the `target_to_llvm_prebuilt` function.

Logging improvements:

* [`crates/rustc_codegen_nvvm/build.rs`](diffhunk://#diff-67ee338edb5803a94f2d3e53ed6d3c3783e2bc4059eaafa6a343a42c8f880548R93): Added a warning message to log the URL of the prebuilt LLVM being used in the `find_llvm_config` function.

Toolchain update:

* [`rust-toolchain`](diffhunk://#diff-335ddbce02d068207f42ecdac988ee267eb014a7a1e7ffe8df7bedd66f716899L8-R8): Updated the Rust toolchain channel from `nightly-2021-12-04` to `nightly-2022-11-22` to use a more recent version.

PR #151, opened 2025-03-08
Panic and emit error when `nvcc` can't be found by `which`

alaric728:
[std::process::Command.output() ](https://doc.rust-lang.org/std/process/struct.Command.html#method.output)doesn't return error on non-zero exit, only when command.inner() fails.

If `which nvcc` can't find `nvcc` it returns like { ExitStatus(-1), "", "" } and causes a panic in the return of the function.

Instead, we should rather panic on non-zero return with an informational diagnostic message.

Seems like this can probably be lumped into PR #143 if we'd like to stuff one PR full of build fixes.
Let me know if we'd rather not panic here and I can see about adapting things to handle the null string.


LegNeato:
Great, thanks for the PR! 🍻

PR #155, opened 2025-03-12
Update toolchain to nightly-03-02.

jorge-ortega:
The purpose of this change is to update the rust toolchain to a more recent release; specifically, `nightly-2025-03-02`.

New toolchain means new rustc internal interface changes that needed to be made. Much of the logic is just copied from the llvm backend, with some changes to conform to nvvm specific logic and where assumptions of opaque pointers deviate with typed pointers. A major refactor to the debug metadata in the llvm backend was also brought in (though likely very unnecessary).

LLVM rust wrapper changes includes:
- backporting opaque pointer specific APIs straight into our wrapper and will be conditionally compiled when using LLVM 7.
- Safer LLVMRust APIs that accept `char*` and `size_t` lengths to create `StringRef`s internally.

Changes to cuda_builder when building cuda targets include:
- building all crates with `no_std`
- building all crates with `saturating_float_casts=false` as nvvm does not support saturating float to int casts.
- registering `nvvm_internal` as a tool attribute in all crates

Complies and runs the `add` example. Neither `path_tracer` or optix examples have been tested with these changes.

Thanks to @apriori for attempting the [previous port](https://github.com/Rust-GPU/Rust-CUDA/pull/98), which served as a helpful guide for this port.


LegNeato:
I'm not qualified to review the LLVM side, the rust backend stuff looks sane. Let's land and iterate as it is strictly better than what is there.

PR #159, opened 2025-03-13
Partial fixes to optix crates and examples.

jorge-ortega:
These changes fixes errors when building `path_tracing_gpu`. It is now able to build and generate ptx. Also migrates a deprecated API in the `denoiser` example.

The host side still fails due to updated dependencies on its side.


LegNeato:
Thanks! 🍻

PR #160, opened 2025-03-15
Fix optix examples

jorge-ortega:
They build and run now!

Screenshot 2025-03-14 203234
Screenshot 2025-03-14 203345

### DIFF
--- a/crates/optix/examples/ex02_pipeline/src/renderer.rs
+++ b/crates/optix/examples/ex02_pipeline/src/renderer.rs
@@ -1,5 +1,5 @@
 use anyhow::{Context, Result};
-use cust::context::{Context as CuContext, ContextFlags};
+use cust::context::Context as CuContext;
 use cust::device::{Device, DeviceAttribute};
 use cust::memory::{CopyDestination, DeviceBox, DeviceBuffer, DevicePointer, DeviceVariable};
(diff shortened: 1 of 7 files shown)


crates/optix/examples/ex02_pipeline/src/renderer.rs
```diff
@@ -40,8 +40,7 @@ impl Renderer {
         let srf_align = device.get_attribute(DeviceAttribute::SurfaceAlignment)?;
         println!("tex align: {}\nsrf align: {}", tex_align, srf_align);
 
-        let cuda_context =
-            CuContext::create_and_push(ContextFlags::SCHED_AUTO | ContextFlags::MAP_HOST, device)?;
```
LegNeato, line 44: Why were these dropped? What are the implications?

jorge-ortega: `create_and_push` comes from the "legacy" Context module, and the modern one didn't have a similar API to create and set flags at the same time. So it just creates whatever the default is, which didn't seem to prevent it from running successfully.

For completeness sakes, we can set flags afterwards to show it can be done in the example.

LegNeato: Nah, I am ok with not showing everything in the example. Thanks!

PR #161, opened 2025-03-15
Output CUDA files in workflow

LegNeato:
This will help debug windows. We'll see how long it takes and how messy it is to see if we actually land it.


LegNeato:
Eh, this is so quick and collapsed by default I'll just land.

Issue #163, opened 2025-03-16
Error compiling example: couldn't load codegen backend symbol


igor-semyonov:
Hi,

I am trying to compile the cpu add example using cargo build.

First I had to uncomment this line: https://github.com/Rust-GPU/Rust-CUDA/blob/1083480841a391a4d8221b10a26628e8b5171abe/crates/rustc_codegen_nvvm/build.rs#L57
This fixed the error of no prebuilt supported LLVM target.

rustc_codegen_nvvm now builds successfully.

Then I did cargo build again on the cpu add example.
I got the following error:
```
    error: couldn't load codegen backend /home/kdcadet/code/Rust-CUDA/target/debug/deps/librustc_codegen_nvvm.so: undefined symbol: _ZN4llvm6Triple9normalizeB5cxx11ENS_9StringRefE
```

Then I checked that so file for that symbol:
```
~ via 🐍 v3.12.9
❯ readelf -Ws /home/kdcadet/code/Rust-CUDA/target/debug/deps/librustc_codegen_nvvm.so|grep _ZN4llvm6Triple9normalizeB5cxx11ENS_9StringRefE
   716: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND _ZN4llvm6Triple9normalizeB5cxx11ENS_9StringRefE
 71968: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND _ZN4llvm6Triple9normalizeB5cxx11ENS_9StringRefE

~ via 🐍 v3.12.9
❯ nm -gD /home/kdcadet/code/Rust-CUDA/target/debug/deps/librustc_codegen_nvvm.so|grep _ZN4llvm6Triple9normalizeB5cxx11ENS_9StringRefE
                 U _ZN4llvm6Triple9normalizeB5cxx11ENS_9StringRefE
```

I am not that familiar with elf or so files, but it seems like the symbol is, in fact, present.

My OS: Gentoo Linux
```
cuda/cpu/add on  main [$!] is 📦 v0.1.0 via 🦀 v1.87.0-nightly took 6s
➜ rustc -Vv
rustc 1.87.0-nightly (8c392966a 2025-03-01)
binary: rustc
commit-hash: 8c392966a013fd8a09e6b78b3c8d6e442bc278e1
commit-date: 2025-03-01
host: x86_64-unknown-linux-gnu
release: 1.87.0-nightly
LLVM version: 20.1.0

[...]

Issue #100, opened 2023-01-28
Support for CUDA 12.0 (libnvvm 2.0)


n4o847:
Hello.

I just upgraded to CUDA 12.0 and I am getting the following error in `cuda_builder`:

```
error: rustc_codegen_nvvm requires at least libnvvm 1.6 (CUDA 11.2)
```

This seems to be because CUDA 12.0 uses libnvvm 2.0, which is incompatible with version 1 [^1] [^2], and the version check fails.

Is there any plan to support CUDA 12.0? Thanks in advance!

### Environment

- Ubuntu 22.04
- Rust nightly-2021-12-04 (the same version as this repository)
- CUDA 12.0
- LLVM 7.1.0

[^1]: https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html
[^2]: https://docs.nvidia.com/cuda/nvvm-ir-spec/index.html

jorge-ortega:
We've just finished updates to a more recent rust toolchain. In my workspace, I've been using 12.8 with libnvvm 2.0, and have changed the line [here](https://github.com/Rust-GPU/Rust-CUDA/blob/45e6b6f7051d8369dba92cbada7763f69335a5ea/crates/rustc_codegen_nvvm/src/nvvm.rs#L66) to bypass that check 🤣

I'm not sure what issues this might cause with either the codegen or the cust/cuda crates today, but for now it seems to "just work".

LegNeato:
Here is the change log: https://docs.nvidia.com/cuda/nvvm-ir-spec/index.html#revision-history

LegNeato:
It doesn't look the libnvvm API was breaking, and I don't think there is any case where we can generate multiple versions concurrently. We don't appear to support the addresspace intrinsics they removed, and there is a reference to addrspacecast in the codebase which is what they suggest using. We don't support pragma unroll, which changed in 2.0.

The only one I am concerned about is:

> Shared variable initialization with non-undef values is no longer supported. In 1.x versions these initializers were ignored silently. This feature makes the 2.0 version incompatible with 1.x versions.

But I doubt it would be an issue unless you are mixing and matching.

So let's closed this as fixed and deal with any issues in follow-up bugs 🚀

PR #165, opened 2025-03-17
Small Compatibility Fix for CUDA 12.8.

Schmiedium:
The latest version of CUDA ships libnvvm with version 2.0, which fails the check for an appropriate version number. This commit fixes this so that 2.0 correctly passes the minimum version number check.

#163 Referencing this issue, this PR should provide a fix for that when using CUDA 12.8

### DIFF
--- a/crates/rustc_codegen_nvvm/src/nvvm.rs
+++ b/crates/rustc_codegen_nvvm/src/nvvm.rs
@@ -63,8 +63,8 @@ pub fn codegen_bitcode_modules(
     // make sure the nvvm version is high enough so users don't get confusing compilation errors.
     let (major, minor) = nvvm::ir_version();
 
-    if minor < 6 && major < 1 {
-        sess.dcx()
+    if major <= 1 && minor < 6 {
+            sess.dcx()
             .fatal("rustc_codegen_nvvm requires at least libnvvm 1.6 (CUDA 11.2)");
     }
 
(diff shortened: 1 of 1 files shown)


LegNeato (changes requested):
I don't think this is right, 1.5 will incorrectly continue / not take the error path. (5 < 6) && (1 < 1) -> (true && false) -> false

LegNeato:
You probably want something like:

```rust
    if major <= 1 && minor < 6
```

LegNeato:
As mentioned in https://github.com/Rust-GPU/Rust-CUDA/issues/100, we are unsure of what the implications here are. I'd be inclined to enable and deal with the fallout though.

Schmiedium:
You're right, I messed up the logic there. Thank you for catching that!

I did some digging and found [here](https://docs.nvidia.com/cuda/nvvm-ir-spec/index.html#revision-history) nvvm version 2.0 is supposed to be a breaking change. I have CUDA 12.8 installed though, and don't see any issues with the examples we have built. 

Everything seems to compile just the same and run fine. If you're okay with the fallout from trying to support newer versions, then I think it would be a good thing to do. I think people trying to use the project in the future will install the newest version of CUDA available, and if this prevents that it could be a headache.

LegNeato:
Right, if you look at #100 I pasted that link in 👍 . It doesn't look the libnvvm API was breaking, and I don't think there is any case where we generate multiple versions concurrently so it is probably fine.

LegNeato (approved):
Thanks!!! 🚀

PR #168, opened 2025-03-17
Add CUDA 12 to CI

LegNeato:
Not sure this is worth the extra time and CPU cycles, we'll see.


LegNeato:
Actually, we should probably drop 11 out of CI as most people will be using 12 as that is what you get from nvidia's download without rooting around.

Issue #133, opened 2025-01-27
Fix CI


LegNeato:
CI is broken and hasn't worked in years. Now that we are rebooting the project (https://github.com/Rust-GPU/Rust-CUDA/issues/130) we need CI.

buk0vec:
Is the plan for CI to still not include testing due to lack of CUDA support in actions? I don't really see another workaround aside from self-hosted runners which I don't think is super feasible on a public repo

LegNeato:
The first step is getting it to build. Then making sure any compiletests work.

At that point I'll rustle up some sponsors for hosted runners (I think they work fine for public repos? Embark had them for rust-gpu)

MikeRomaniuk:
For the CI there is a great tool: https://github.com/matklad/cargo-xtask
I'd be eager to see this adopted in this project.

Schmiedium:
I've made some progress on this, I updated CI in my own fork so that it gets far enough to build the project, at least on the Ubuntu container. I've made some progress towards getting the project to build, there were a lot of places where the codegen no longer adheres to the rustc private api, and I think I've fixed and updated a lot of those issues. 

I'm going to look into the rust-gpu codegen crate to see what that looks like and hopefully gain some insight that'll help me fix the nvvm codegen

LegNeato:
@Schmiedium I merged your PR, but I think the `rustc` version is going to be much too new to actually work.

Schmiedium:
Thank you! That makes sense, I wasn't quite sure which version to target here. Do you have a rough idea of which version would be good to try and target? I can try to use the same one as Rust-GPU, or do you think that's still too new?

LegNeato:
Rust-gpu had to vendor in parts of the compiler to get to that version. I would do the one before, as that's the one I personally forward ported and it wasn't too hard. I'm not a computer right now, but you should be able to check to get history

Schmiedium:
Excellent, I'll target that then. Thanks!

LegNeato:
I don't use windows, anyone want to take a crack at fixing the windows CI link error? Comment here if so!

Schmiedium:
I can take a crack at it, I think it should just be some environment variables to set properly. I was planning to put together a windows container with CUDA pre installed to hopefully speed some things up too, so if that works it'll hopefully speed up and fix the windows CI

jorge-ortega:
[...]

PR #169, opened 2025-03-17
Mark functions in float_ext.rs as gpu_only

Schmiedium:
Different behavior between linux and windows linkers causes windows CI to fail when compiliing crates that need cuda_std.

[Microsoft provides this explanation for the linker error.](https://learn.microsoft.com/en-us/cpp/error-messages/tool-errors/linker-tools-error-lnk2019?view=msvc-170#you-call-intrinsic-functions-or-pass-argument-types-to-intrinsic-functions-that-arent-supported-on-your-target-architecture)

This link contains the exact reason for the error message. Compiling float_ext for an architecture other than nvptx caused the linker to fail with LNK2019, for the reason linked.

#133 See this issue

Thanks to @jorge-ortega for figuring this out.

### DIFF
--- a/crates/cuda_std/src/float_ext.rs
+++ b/crates/cuda_std/src/float_ext.rs
@@ -1,7 +1,7 @@
 //! Extension trait for [`f32`] and [`f64`], providing high level wrappers on top of
 //! raw libdevice intrinsics from [`intrinsics`](crate::intrinsics).
 
-use crate::intrinsics as raw;
+use cuda_std_macros::gpu_only;
 
 // allows us to add new functions to the trait at any time without needing a new major version.
 mod private {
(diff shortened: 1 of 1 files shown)


Schmiedium:
@LegNeato I see you're working on some new CI changes, do you want me to rebase this on top of that when you're done?

crates/cuda_std/src/float_ext.rs
```diff
@@ -1,7 +1,7 @@
 //! Extension trait for [`f32`] and [`f64`], providing high level wrappers on top of
 //! raw libdevice intrinsics from [`intrinsics`](crate::intrinsics).
 
-use crate::intrinsics as raw;
```
jorge-ortega, line 4: Cuda target will fail to build without this when it actually tries to call `raw::cospi`, etc., below. Gate this with cfg, or replace all the alias with the full path.

jorge-ortega (changes requested):

LegNeato:
@Schmiedium nah, it shouldn't matter as my changes aren't touching what you are doing (linux only)

crates/cuda_std/src/float_ext.rs
Schmiedium: Understood, thanks! Should the gate be just for the nvptx64 architecture, or should I also specify target_os?

jorge-ortega: They both mean the same thing today. Targeting just the architecture should be enough.

jorge-ortega (approved):

PR #162, opened 2025-03-15
Fix examples

trigpolynom:
built successfully based on changes made on top of the work @jorge-ortega has already done in #160. Below are some config params I had to set for my Ubuntu 22.04 env:

# Set CUDA Root and Paths
export CUDA_ROOT=/usr/lib/cuda
export CUDA_PATH=$CUDA_ROOT
export CUDA_NVVM_ROOT=$CUDA_ROOT/nvvm

# CUDA Binary and Library Paths
export PATH=$CUDA_ROOT/bin:$CUDA_ROOT/nvvm/lib64:/root/.cargo/bin:$PATH
export LD_LIBRARY_PATH=$CUDA_ROOT/lib64:$CUDA_ROOT/nvvm/lib64:$CUDA_ROOT/nvvm/libdevice:$LD_LIBRARY_PATH
export LIBRARY_PATH=$CUDA_ROOT/lib64:$CUDA_ROOT/nvvm/libdevice:$LIBRARY_PATH

# Rust and OptiX Configuration
export LLVM_LINK_STATIC=1
export RUST_LOG=info
export OPTIX_ROOT=$HOME/optix/
export OPTIX_ROOT_DIR=$OPTIX_ROOT

# Add Rust Toolchain Path
export PATH=$HOME/.cargo/bin:$PATH
export LD_LIBRARY_PATH=$HOME/.rustup/toolchains/nightly-2025-03-02-x86_64-unknown-linux-gnu/lib:$LD_LIBRARY_PATH
export RUSTFLAGS="-C link-args=-ltinfo"

PR #171, opened 2025-03-19
Fix path_tracer example

jorge-ortega:
Originally #162 

Fixes path_tracer example to build and run. Still some runtime errors when closing the crab ui, and can sometime AV when closing the app.


LegNeato:
Hmmm, I know these require another install but I wonder if that is the only reason they are disabled in CI. I'll file a followup to make a job for this, otherwise it will regress.

Thanks @trigpolynom and @jorge-ortega 🍻

PR #174, opened 2025-03-20
Rework optix_stubs.c

jorge-ortega:
Doing a diff of optix_stubs.c in tree with optix_stubs.h from the SDK showed only whitespace differences and removal of `inline` from all the stubs so they end up in the binary when compiling them. This change instead includes optix_stubs.h and uses the preprocessor to erase `inline` in the header. 

This should make it easier to support newer optix versions without having to update the stubs to match.

Edit: CI won't build this, but this does still build and run the path_tracer example for me on Windows.


LegNeato:
Learned about `CargoCallbacks`, neat! Nice and straightforward PR 👍

Issue #179, opened 2025-03-30
Consider switching from `vek` to `glam`


LegNeato:
I can't find any backstory for using `vek` over another lib.

`rust-gpu` uses `glam` and `rust-cuda` uses `vek`. As far as I can see, there is no real reason for it being different. It would be nice to use the same so we don't have to do import `cfg()` dance when sharing code between vulkan and cuda.

`cust` / the host side has a feature for `impl_glam`, but it looks on the GPU side we just support vek in cuda std?

Note glam has this feature:

* `cuda` - forces `glam` types to match expected [cuda alignment]

Advantages for `glam`:

* popular
* unopinionated, doesn't really define anything for you except column-major matrix order
* one of the fastest vector libs (for CPU), the dev also runs the [mathbench](https://github.com/bitshifter/mathbench-rs) to compare performance against other crates
* has no_std support via libm (which vek does too)
* glam also has support for the rkyv serializer 

(thanks to @Firestar99 for the list).

PR #180, opened 2025-04-01
Replace `vek` with glam.

LegNeato:
This fixes https://github.com/Rust-GPU/Rust-CUDA/issues/179.


jorge-ortega:
Seems like we're resolving to multiple versions of glam between what's re-exported through cust and what's pulled it in directly in some places. We might want to consider using workspace dependencies to ensure we use the same version across our crates.

LegNeato:
Yeah, I just did some dumb find and replaces 😄 . I intend to actually look at this in a bit.

adamcavendish:
I personally think that we don't need vek nor glam for dim1,dim2,dim3 conversion. Mostly users won't use a vek nor a glam vectorized type for creating kernel invocations. To use glam or vek, people should add that as a dependency to their kernel projects instead.

LegNeato:
I don't quite know the history, but rust-gpu didn't have it and it was specifically added. There likely was a reason vek was added here as well.

jorge-ortega:
After taking a closer look, I do find the thread/index functions (`index_{2,3}d`, `{block,thread}_idx`, etc.) returning a specific vector type of choice a bit of an overstep. vek and glam both implement `From` for tuples and arrays. It would better to see these functions return either one of those so users can call `into` to convert to the vector type of their choice. This would also avoid imposing a vector type dependency in cuda_std.

LegNeato:
I think ideally we'd have a type or trait in `core` that SIMD and GPUs (and wasm SIMD) can all use.

I've also been waiting for [externally implementable items](https://github.com/rust-lang/rust/issues/125418) to land, as maybe we just define our own trait and let end-user code plug into it, similar to allocators and panic handlers.

jorge-ortega:
I have a [branch](https://github.com/jorge-ortega/Rust-CUDA/tree/glam) of this on my fork that's rebased on main and update to fix the remaining errors.

LegNeato:
Cool, I can close this one out. Sorry, have had some personal stuff I've had to deal with lately so I've been a bit busy. Hoping it gets cleared up at the start of May!

Issue #166, opened 2025-03-17
Regenerate `cust_raw`for newer CUDA / support multiple versions


LegNeato:
Now that we are adding support for newer CUDA (#100), we need a way to select the version for `cust_raw`'s bindgen (and elsewhere?) See: https://github.com/Rust-GPU/Rust-CUDA/blob/400f816f382843559175ac8df3a46d5858188bd7/crates/cust_raw/README.md?plain=1#L6

jorge-ortega:
Currently we have a checked-in generated bindings that seem to be for windows based on the `repr(i32)` value for enums. That probably won't break linux, and we could update current binding and also generate linux bindings, but neither seems not ideal. 

In the usual bindgen flow, the user selects the version through whatever they have installed on their machine, and cust_raw (typically -sys) crate would generate bindings for that version. The cust crate can then configure itself through reading the version used to generate the bindings.

Say we have support up to 12.8, and the user also has 12.0 installed. cust _should_ be able to support APIs up to 12.0 and disable newer once through its configuration. If the user has a newer version, then at best cust could still compile without any support for newer features/APIs, but the user _should_ still be able to access them through the sys crate. At worst, we don't compile due to some backwards incompatible change with the newer version.

LegNeato:
Right, we need to decide if we generate and check in plus the hide the variants behind targets + features or we generate on the fly via `build.rs` and what is installed. The former is sort of gross but we see the complete picture and changes for each version and have more visibility when bugs are filed. The later means cross compilation is unsupported.

jorge-ortega:
I gave generating bindings at build time a shoot and have a WIP [branch](https://github.com/jorge-ortega/Rust-CUDA/blob/cuda-versions-v12) focusing on cuda 12 and up for now on Windows. I built with versions 12.0, 12.1, 12.3, and 12.8. The only breaking changes were in the graph api. 12.3 introduced conditional nodes, so cust checks support for that and sets a cfg so the wrapper enum can account for it. With these changes, cust is able to build for on every version I tested. 12.0 can't codegen thought, as libnvvm doesn't like our data layout for some reason, but 12.1 and up were ok with it.

adamcavendish:
[...]

PR #187, opened 2025-04-03
refactor(cust_raw): Parse macro function renames in cuda headers.

jorge-ortega:
Adds a custom bindgen callback that prevents function renames due to macro defines while still linking to the intend function after macro expansion. It does this by tracking macros when generating the bindings, so that it can change the name of the function back to what it was before the macro changed and link to the macro expanded function name. Doing so helps prevents breaking changes across CUDA versions when generating the bindings, and generates function bindings that match those used in Nvidia's CUDA documentation.

misc.: optix-sys rebuilds if related optix environment variables change.
misc.: unpin cc and jobserver from add example.


jorge-ortega:
TBH I don't know why Windows with v12 is passing when we never set LIBCLANG_PATH or installed Clang...

adamcavendish:
> TBH I don't know why Windows with v12 is passing when we never set LIBCLANG_PATH or installed Clang...

Maybe it uses visual studio toolchains?

adamcavendish:
A heads-up that I'm also working on a change to use prebuilt container images that contains NVidia toolchains so I might conflict with your github actions changes.

jorge-ortega:
Leaving this open in case anyone has comments and will merge in the morning unless @LegNeato beats me to it. I'll have additional changes to add move coverage of the sdk to try match what cudarc offers.

jorge-ortega:
Linking #166

PR #188, opened 2025-04-03
feat(ci): Split Linux CI workflow and add container images pipeline

### LINKED ISSUE #185: Switch CI to use containers
GitHub Actions appears to support running steps in containers:

https://docs.github.com/en/actions/writing-workflows/choosing-where-your-workflow-runs/running-jobs-in-a-container

It's probably good to just reuse the public NVIDIA containers and/or our Dockerfiles in the repo so that everything is installed by nvidia. This will, for example, fix the fact that cudnn is not installed as they have containers with it. We can focus our efforts on the rust-specific stuff.

NVIDIA's [supported container tags](https://gitlab.com/nvidia/container-images/cuda/blob/master/doc/supported-tags.md) show they support ubuntu 20 and 24 with the CUDA versions we need.

@adamcavendish already did work creating Dockerfiles in https://github.com/Rust-GPU/Rust-CUDA/tree/main/container. Ideally we would have one set that we use for CI and point people to.

adamcavendish:
Added a new workflow, container_images.yml, to automate building and pushing CI container images to GitHub Container Registry, ensuring consistent environments for CI runs across Linux distributions.

Added a dedicated CI workflow for Linux to streamline testing across multiple CUDA and OS configurations. It reuses the images built from container_images.yml workflow to avoid installing CUDA toolkits every time to save time.

Key changes:
- Matrix configurations for Ubuntu 22/24, RockyLinux 9 with CUDA 12.8.1
- Dockerfiles updated with required dependencies (clang, CUDA toolchains, etc.) to support bindgen and build examples.
- Removed redundant CI steps from `rust.yml` now handled in `ci_linux.yml`

----

Resolves #185.

Please note that I have disabled `push` on `pull_request` events so we need to manually trigger once before we can have a successful Linux CI run.


adamcavendish:
@LegNeato As mentioned in the description, would you help trigger the `Build CI Container Images` workflow manually so the CI can correctly use these images?

jorge-ortega:
I think there is a permission issue because the branch is from a fork. I'm going to duplicate the branch on the main repo which should kick off the container imagine builds with the right permissions.

jorge-ortega:
That seemed to do the trick. Going to merge and rebase the bindgen changes. Thanks @adamcavendish!

adamcavendish:
> I think there is a permission issue because the branch is from a fork. I'm going to duplicate the branch on the main repo which should kick off the container imagine builds with the right permissions.

Yeah, that's why I posted a link showing that my own actions are working correctly so you guys will be confident to take it.

PR #189, opened 2025-04-03
chore(ci) Use CUDA network installer and specify sub-packages.

jorge-ortega:
The network installer is a smaller initial download, which will only download the require packages to install. This should reduce the install times on Windows CI.


jorge-ortega:
Install times are down to 3 minutes on Windows. 😏

@adamcavendish I have your branch rebased on this and will try to update your PR after I merge.

PR #190, opened 2025-04-05
chore(example): add nvvm library path to LD_LIBRARY_PATH and add back the `add` example in CI

### LINKED ISSUE #129: Running examples: Dependency problems, Linux target, errors in codegen_nvvm
System: Alma Linux 8.10
CUDA: 12.6.3

I cloned the Rust-CUDA and tried to compile `examples/cuda/cpu/add` using `cargo run`, having Rust nightly installed. The `rust-toolchain` file makes rustup install 1.59.0-nightly. However, the `log`-crate, which is required in version 0.4.25, needs Rust 1.60.0 or higher.

Changing the toolchain `nightly-2022-04-07` leads to a similar error, where `openssl-sys v0.9.104` needs 1.63.0 or higher.

Then, I removed the toolchain-file to just use the newest version of the rustc. The problem with the dependencies was solved, but the buildscript of `rustc_codegen_nvvm` nags about the target. In the buildscript I found this uncommented match arm

```
// NOTE(RDambrosio016): currently disabled because of weird issues with segfaults and building the C++ shim
// "x86_64-unknown-linux-gnu" => "linux-x86_64",
```

**Is Linux basically not supported?**

I removed the comment and tried again, this time getting a bunch of errors while trying to compile codegen_nvvm. Some of them where solved running `rustup component add rust-src rustc-dev llvm-tools-preview`. However, I'm stuck here now:

1. I still get the error that the `extern crate rustc_attr` could not be found, and the suggested solution is again to install the llvm tools, which I already did.
2. More importantly, I get 104 other errors like `use rustc_codegen_ssa::traits::BaseTypeMethods;: no BaseTypeMethods in traits` in `crates/rustc_codegen_nvvm/src/abi.rs:9:5`, which I have no idea how to solve.

Does anyone know how to resolve the issues?

adamcavendish:
When linking `libnvvm.so.4` into `librustc_codegen_nvvm.so` shared library via the `cust_raw` package, we specified the following parameters in the `build.rs`:

- cargo::rustc-link-search=native=xxx
- cargo::rustc-link-lib=dylib=nvvm

According to the [Cargo documentation](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-link-search), the path we added to the `rustc-link-search` parameter is not further populated to the `LD_LIBRARY_PATH` for search by the transitive dependencies.

> These paths are also added to the dynamic library search path environment variable if they are within the OUT_DIR.

The behavior is intended to be limited to the OUT_DIR so the `libnvvm.so.4`, which lives as a part of the CUDA SDK, is excluded. Therefore, we must add it to the LD_LIBRARY_PATH manually in our container build stage.

----

Since this PR changes the container, we would need to retrigger a container build manully before we can see correct CI results.

Resolves #129.

### DIFF
--- a/.github/workflows/ci_linux.yml
+++ b/.github/workflows/ci_linux.yml
@@ -44,10 +44,10 @@ jobs:
       - name: Clippy
         env:
           RUSTFLAGS: -Dwarnings
-        run: cargo clippy --workspace --exclude "optix*" --exclude "path_tracer" --exclude "denoiser" --exclude "add" --exclude "ex*" --exclude "cudnn*"
+        run: cargo clippy --workspace --exclude "optix*" --exclude "path_tracer" --exclude "denoiser" --exclude "ex*" --exclude "cudnn*"
       - name: Build
-        run: cargo build --workspace --exclude "optix*" --exclude "path_tracer" --exclude "denoiser" --exclude "add" --exclude "ex*" --exclude "cudnn*"
+        run: cargo build --workspace --exclude "optix*" --exclude "path_tracer" --exclude "denoiser" --exclude "ex*" --exclude "cudnn*"
       - name: Check documentation
         env:
           RUSTDOCFLAGS: -Dwarnings
-        run: cargo doc --workspace --all-features --document-private-items --no-deps --exclude "optix*" --exclude "path_tracer" --exclude "denoiser" --exclude "add" --exclude "ex*" --exclude "cudnn*" --exclude "cust_raw"
+        run: cargo doc --workspace --all-features --document-private-items --no-deps --exclude "optix*" --exclude "path_tracer" --exclude "denoiser" --exclude "ex*" --exclude "cudnn*" --exclude "cust_raw"
--- a/container/rockylinux9-cuda12/Dockerfile
+++ b/container/rockylinux9-cuda12/Dockerfile
@@ -38,5 +38,8 @@ WORKDIR /data/Rust-CUDA
 RUN rustup show
 RUN rm -f "rust-toolchain.toml"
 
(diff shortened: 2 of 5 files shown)


adamcavendish:
@jorge-ortega In addition, should we make the develop container packages public in the rust-gpu org so other people can download and use them if they want?

jorge-ortega (approved):

jorge-ortega:
I think that might be a question for @LegNeato 

Any idea why setting the LD_LIBRARY_PATH in container didn't seem to have an effect?

adamcavendish:
> I think that might be a question for @LegNeato 
> 
> Any idea why setting the LD_LIBRARY_PATH and running `ldconfig` in container didn't seem to have an effect?

Linux env must be exported before it can take effect. I did test via export LD_LIBRARY_PATH in container and it did work. ld.conf has a cache as well so it must be updated to take effect too.

jorge-ortega:
I thought I tried `export LD_LIBRARY_PATH=...` and `echo "/libvvm/path" >> /etc/ld.so.conf && ldconfig` with no effect. I'll just leave it as a me problem. Thanks for covering this!

LegNeato:
> @jorge-ortega In addition, should we make the develop container packages public in the rust-gpu org so other people can download and use them if they want?

Yeah, I think that would be better rather than people having to figure out how to cobble together a build env.

LegNeato:
> > @jorge-ortega In addition, should we make the develop container packages public in the rust-gpu org so other people can download and use them if they want?
> 
> Yeah, I think that would be better rather than people having to figure out how to cobble together a build env.

@adamcavendish I flipped these settings, the container packages should now be public.

PR #191, opened 2025-04-07
fix(cust_raw): Use string macro expansion to record function renames

jorge-ortega:
`cl.exe` macro expansion does not do any token validation when using token pasting (##),  but `cc` does validate token pasting results in a valid token. This change switches from token pasting to stringify (#), which expands all the macro function variables as strings. Rename lines now  expand to `"RUST_RENAME" "_someDefine" "expandedFuncName"`. This prevents `cc` from erroring during expanding in cases where the expanded macro would have resulted in an invalid token, i.e: `RUST_RENAME_someDefine_(not_valid)` which is invalid because the `(` is not a valid token character.


adamcavendish:
When we test the cust_raw pkg, we might need to enable all features so all the bindgens are generated.

jorge-ortega:
Added a CI build step to do just that. I learned Windows mounts remain case insensitive in a dev container.

adamcavendish:
> Added a CI build step to do just that. I learned Windows mounts remain case insensitive in a dev container.

Yeah, should be same across case insensitive filesystems.

PR #198, opened 2025-04-14
Rename and retitle rust workflow to be windows workflow

LegNeato:
We now have a different workflow for linux, so let's make this clearer.


adamcavendish:
That's awesome.

Issue #182, opened 2025-04-01
Ownership of `cuda-*-sys` crates


LegNeato:
Hello @termoshtt ! We saw [your post on reddit](https://www.reddit.com/r/rust/comments/1jmh4af/rustcudacudasys_has_been_archived/), thank you for pointing people to rust-cuda 🎉 . We're currently [revamping our `-sys` crate organization](https://github.com/Rust-GPU/Rust-CUDA/pull/181) and it would be awesome if we could use the `cuda-*-sys` crate names. If you are willing, feel free to just transfer ownership to me (@LegNeato) in crates.io. We would of course bump the version on our next release so people don't accidentally switch. Thanks! 🍻

LegNeato:
Hey @termoshtt , just wanted to make sure you saw this 👋

PR #201, opened 2025-04-15
reactor(cust_raw): Generate bindings for cuda types and restructure `cust_raw`'s crates

jorge-ortega:
- Allow list files instead of types/var/functions.
- Split out type headers from runtime/cublas to their own crate.
- Drop sys prefix from internal crates. (Hopefully we can still get ownership of cuda-sys, which would make the internal sys suffix redundant).

This restructures cust_raw to the following:
```
src/
├── cublas/        # Cublas and its extensions
│   ├── core.rs
│   ├── lt.rs
│   ├── mod.rs
│   └── xt.rs
├── types/         # Common CUDA Runtime types
│   ├── driver.rs
│   ├── library.rs
│   ├── mod.rs
│   ├── surface.rs
│   └── texture.rs
├── driver.rs
├── lib.rs
├── nvptx.rs
├── nvvm.rs
├── runtime.rs
```
This is a precursor to adding other CUDA library bindings, like nvJPEG and cudaFFT.

### DIFF
--- a/crates/blastoff/src/context.rs
+++ b/crates/blastoff/src/context.rs
@@ -4,8 +4,8 @@ use std::os::raw::c_char;
 use std::ptr;
 
 use cust::stream::Stream;
-use cust_raw::cublas_sys;
-use cust_raw::driver_sys;
+use cust_raw::cublas;
+use cust_raw::driver;
 
 use super::error::DropResult;
 use super::error::ToResult as _;
@@ -73,7 +73,7 @@ bitflags::bitflags! {
 /// - [Matrix Multiplication <span style="float:right;">`gemm`</span>](CublasContext::gemm)
 #[derive(Debug)]
 pub struct CublasContext {
-    pub(crate) raw: cublas_sys::cublasHandle_t,
+    pub(crate) raw: cublas::cublasHandle_t,
 }
 
 impl CublasContext {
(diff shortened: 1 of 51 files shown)


adamcavendish:
The `_sys` suffix was originally used because it was a bindgen generated system bindings. I propose that we should align with the community and call them `-sys` packages.

jorge-ortega:
The convention tends to only apply to the top-level crate name, not the crates inside of it. The raw in cust_raw _kinda_ conveys what sys is supposed to (there be bindings in here), so if you consider that it comes off a bit redundant:
```
import cust_raw::driver_sys
            ^           ^_bindings, dammit!
            |_bindings
```
The important thing is users understand that the crate as a whole only contains bindings before they even add it as a dependency, which should make them gravitate to the higher-level crates. 

Still hoping on hearing back from the owner of cuda-sys in #182.

adamcavendish:
> The convention tends to only apply to the top-level crate name, not the crates inside of it. The raw in cust_raw _kinda_ conveys what sys is supposed to (there be bindings in here), so if you consider that it comes off a bit redundant:
> 
> ```
> import cust_raw::driver_sys
>             ^           ^_bindings, dammit!
>             |_bindings
> ```
> 
> The important thing is users understand that the crate as a whole only contains bindings before they even add it as a dependency, which should make them gravitate to the higher-level crates.
> 
> Still hoping on hearing back from the owner of cuda-sys in #182.

Yeah, if we do can get the cuda-sys crate, I agree that we can drop the "-sys" suffix.

LegNeato:
I reached out to them over email.

jorge-ortega:
I'm splitting out the module renames from this change and will open a separate PR for that. This should now contain just the restructuring and new type bindings.

Clippy seems to be building the examples and is now failing during LTO. I don't repro this on windows or in ubuntu24 devcontainer.

PR #206, opened 2025-04-26
fix(rustc_codegen_nvvm): Check function signatures before overriding libm functions.

### LINKED ISSUE #205: Latest libm 0.2.13 is failing during LTO
As of a few days ago, our kernel crates fail to build in CI with the following error (plus some additional logging I added)
```
error: failed to parse bitcode for LTO module libm.9303cbbb082352b3-cgu.09: Invalid cast (Producer: 'LLVM7.0.1' Reader: 'LLVM 7.0.1')
```
Turns out there was a new release for libm on April 22. I originally wasn't able to repro this, as the Cargo.lock file I had was set to libm 0.2.11.  After regenerating my lock file, which updates to libm 0.2.13, I'm able to repro this in on windows and ubuntu 24. 

I have yet to figure out what changed in libm that's causing the codegen to fail during LTO. In the meantime, the best work around seems to be lock libm to 0.2.11:

```toml
[[package]]
name = "libm"
version = "0.2.11"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "8355be11b20d696c8f18f6cc018c4e372165b1fa8126cef092399c9951984ffa"
```

jorge-ortega:
libm 0.2.13 refactored some of its implementations to use generic functions. The monomorphized function names clashed with the function names the codegen would override, causing it the codegen to override the function to a libdevice call that did not match the function signature. It then tries to patch this with a bitcast betwen floating point types, which are invalid casts.

This change adds an additional check before overriding. The libm function signature must match the libdevice intrinsic signature of the same name. Note that this prevents overrides even if bitcasts between the two argument types would be valid, which is likely another bug anyways.

fixes #205

### DIFF
--- a/crates/rustc_codegen_nvvm/src/override_fns.rs
+++ b/crates/rustc_codegen_nvvm/src/override_fns.rs
@@ -2,11 +2,16 @@
 //! codegen-builtin methods. Currently the only use for this is overriding libm functions
 //! with libdevice intrinsics (which are much faster and smaller).
 
-use crate::{builder::Builder, context::CodegenCx, llvm};
+use crate::abi::FnAbiLlvmExt;
+use crate::builder::Builder;
+use crate::context::CodegenCx;
+use crate::llvm;
 use rustc_codegen_ssa::mono_item::MonoItemExt;
-use rustc_codegen_ssa::traits::BuilderMethods;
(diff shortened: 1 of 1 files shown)


LegNeato:
This hit rust-gpu as well: https://github.com/Rust-GPU/rust-gpu/issues/242

PR #213, opened 2025-05-13
build(cudnn-sys): Add CUDNN_INCLUDE_DIR

giantcow:
Enables users to specify a non-standard cuDNN install path. This seems to be needed for the newer editions of the CUDA toolkit, as cuDNN isn't included by default (at least in the Fedora repo's, you have to install from a tarball)

### DIFF
--- a/crates/cudnn-sys/build/cudnn_sdk.rs
+++ b/crates/cudnn-sys/build/cudnn_sdk.rs
@@ -2,6 +2,7 @@ use std::env;
 use std::error;
 use std::fs;
 use std::path;
+use std::path::Path;
 
 /// Represents the cuDNN SDK installation.
 #[derive(Debug, Clone)]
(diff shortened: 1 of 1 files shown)


LegNeato:
Thanks for the PR!

I think it would be more idiomatic like:

```rust
let mut cudnn_paths: Vec<&Path> = CUDNN_DEFAULT_PATHS.iter().map(Path::new).collect();
if let Some(override_path) = env::var_os("CUDNN_INCLUDE_DIR") {
    cudnn_paths.push(Path::new(&override_path));
}

cudnn_paths
    .iter()
    .find(|p| Self::is_cudnn_include_path(p))
    .map(|p| p.to_path_buf())
    .ok_or("Cannot find cuDNN include directory.".into())
```

LegNeato (changes requested):
Thanks for the PR and welcome to the project 🍻 !

I think it would be a bit more idiomatic like:

```rust
let mut cudnn_paths: Vec<&Path> = CUDNN_DEFAULT_PATHS.iter().map(Path::new).collect();
if let Some(override_path) = env::var_os("CUDNN_INCLUDE_DIR") {
    cudnn_paths.push(Path::new(&override_path));
}

cudnn_paths
    .iter()
    .find(|p| Self::is_cudnn_include_path(p))
    .map(|p| p.to_path_buf())
    .ok_or("Cannot find cuDNN include directory.".into())
```

LegNeato:
Thanks again for the PR! 🥳

CharryWu:
This env var is useful for me to supply the latest v13.0 CUDA include path to enable build for cudnn-sys module. The hardcoded windows system default include path is outdated and only supports cuda v8 v9

> Enables users to specify a non-standard cuDNN install path. This seems to be needed for the newer editions of the CUDA toolkit, as cuDNN isn't included by default (at least in the Fedora repo's, you have to install from a tarball)

PR #223, opened 2025-06-01
Fix read_volatile intrinsic

brandonros:
rebased version of https://github.com/Rust-GPU/Rust-CUDA/pull/216


LegNeato:
I just rebased https://github.com/Rust-GPU/Rust-CUDA/pull/216 so this can probably be closed.

Issue #253, opened 2025-08-21
Adding support for standard atomic intrinsics


FractalFir:
Currently, `Rust Cuda` does not support the standard atomic intrinsics, used by `core` & `std`. Instead, `Rust Cuda` provides its own set of atomics in `cuda_std`. 

From reading the codebase, the primary issue seems to be this:
https://github.com/Rust-GPU/Rust-CUDA/blob/33664c024336de23b7e912a56b9888f6c5708b2f/crates/rustc_codegen_nvvm/src/builder.rs#L1135
1. CUDA only supports 32 bit, 64 bit and 128 bit atomics(no 8 & 16 bit) AND
2. CUDA atomics have some address-space restrictions(they only work in the `global` and `shared` address space).  

Both of those problems can be solved(or at least heavily mitigated), and at a very tiny performance cost. 

# Address spaces

Ignoring the `generic` address space(which is just a superset of all the address spaces), CUDA has 4 address spaces: `global`,`shared`,`constant`, `local`. 

The `cmpxchg` instruction(and other atomics, like `atomicrmw`) work fine with `global` & `shared` : the docs say the pointer must be in one of those address spaces.

https://docs.nvidia.com/cuda/nvvm-ir-spec/index.html#cmpxchg-instruction

This means that the instruction will not work with data in the `constant` or `local` address space. This seems to be the reason why this instruction is not used to implement atomics in Rust-CUDA.

## Constant address space: UB in Rust anyway. 

We can ignore the `constant` address space: Using *any* [atomics on constant data is UB in Rust anyway,](https://doc.rust-lang.org/std/sync/atomic/#atomic-accesses-to-read-only-memory) so we are free to assume no sound code uses atomics with the constant address space. 

With this, only `local` the address space remains. 

## Local address space

Local address space is a bit tricky. I *assume* it is illegal for anything but the current thread to access the local address space. If this is the case, we can use non-atomic instructions here, and soundly emulate atomics. 

## Putting this all together:

Now, we have figured out what to do for each address space. We can use this info to make the right choice *based on the address space a pointer belongs to*. 

Thankfully, NVVM IR provides us with helpfull intrinsics to do so: 
```
    i1 @llvm.nvvm.isspacep.const(i8*)
    i1 @llvm.nvvm.isspacep.global(i8*)
    i1 @llvm.nvvm.isspacep.local(i8*)
    i1 @llvm.nvvm.isspacep.shared(i8*)
```
[...]

PR #263, opened 2025-09-11
Add support for standard Rust atomic compare exchange intrinsics.

FractalFir:
Currently, Rust-CUDA only supports CUDA-specifc atomics, and does not support the atomic operations provided by the Rust standard library. 

This PR aims to address that, and emulate those intrinsics(by using CUDA-specific ones). The fix I implement is described in: https://github.com/Rust-GPU/rust-cuda/issues/253

Once this PR(with the basic infrastructure for intrinsics) gets merged, we can add support for all the other intrinsics, allowing a lot more Rust code to run on the GPU - out of the box.

### DIFF
--- a/crates/rustc_codegen_nvvm/src/builder.rs
+++ b/crates/rustc_codegen_nvvm/src/builder.rs
@@ -1115,15 +1115,36 @@ impl<'ll, 'tcx, 'a> BuilderMethods<'a, 'tcx> for Builder<'a, 'll, 'tcx> {
     ) -> (&'ll Value, &'ll Value) {
         // LLVM verifier rejects cases where the `failure_order` is stronger than `order`
         match (order, failure_order) {
-            (AtomicOrdering::SeqCst, _) => (),
-            (_, AtomicOrdering::Relaxed) => (),
-            (AtomicOrdering::Release, AtomicOrdering::Release)
-            | (AtomicOrdering::Release, AtomicOrdering::Acquire)
+            // Failure order `Release` & `AcqRel` is simply invalid.
+            (_, AtomicOrdering::Release | AtomicOrdering::AcqRel) => {
+                self.abort();
+                return (
+                    self.const_undef(self.val_ty(cmp)),
+                    self.const_undef(self.type_i1()),
+                );
+            }
+            // Success & failure ordering are the same - OK.
+            (AtomicOrdering::SeqCst, AtomicOrdering::SeqCst)
+            | (AtomicOrdering::Relaxed, AtomicOrdering::Relaxed)
             | (AtomicOrdering::Acquire, AtomicOrdering::Acquire) => (),
+            // Failure is `SeqCst`(strongest) & success is anything else(weaker) - reject.
+            (_, AtomicOrdering::SeqCst) => {
+                self.abort();
+                return (
+                    self.const_undef(self.val_ty(cmp)),
+                    self.const_undef(self.type_i1()),
+                );
+            }
+            // Failure is Relaxed(weakest), and success is anything - OK.
+            (_, AtomicOrdering::Relaxed) => (),
+            // Failure is anything, and success is SeqCest(strongest) - OK.
+            (AtomicOrdering::SeqCst, _) => (),
+            // Failure is Acquire, and success is Release - OK.
+            (AtomicOrdering::Release, AtomicOrdering::Acquire) => (),
+            // Success is AcqRel & failure is Acquire - OK
             (AtomicOrdering::AcqRel, AtomicOrdering::Acquire) => (),
-            (AtomicOrdering::Relaxed, _)
-            | (_, AtomicOrdering::Release | AtomicOrdering::AcqRel | AtomicOrdering::SeqCst) => {
-                // Invalid cmpxchg - `failure_order` is stronger than `order`! So, we abort.
+            // Success is weaker than failure - reject.
+            (AtomicOrdering::Relaxed, AtomicOrdering::Acquire) => {
                 self.abort();
                 return (
                     self.const_undef(self.val_ty(cmp)),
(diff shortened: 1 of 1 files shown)


FractalFir:
Now, we should support *almost* all atomics(including loads and stores), besides atomic `nand`, which is rarely used(and pretty straightforward to emulate).

crates/rustc_codegen_nvvm/src/builder.rs
```diff
@@ -1134,26 +1106,126 @@ impl<'ll, 'tcx, 'a> BuilderMethods<'a, 'tcx> for Builder<'a, 'll, 'tcx> {
...
+        match (order,failure_order){
+            (AtomicOrdering::SeqCst, _)=>(),
+            (_, AtomicOrdering::Relaxed)=>(),
+            (AtomicOrdering::Release, AtomicOrdering::Release) |  (AtomicOrdering::Release, AtomicOrdering::Acquire) | (AtomicOrdering::Acquire, AtomicOrdering::Acquire)=>(),
+            (AtomicOrdering::AcqRel,AtomicOrdering::Acquire)  => (),
+            (AtomicOrdering::Relaxed, _) | (_, AtomicOrdering::Release | AtomicOrdering::AcqRel | AtomicOrdering::SeqCst)=>{
+                // Invalid cmpxchg - `failure_order` is stronger than `order`! So, we abort.
+                self.abort();
+                return (self.const_undef(self.val_ty(cmp)),self.const_undef(self.type_i1()));
+            }
```
LegNeato, line 1126: Your match accepts (Release, Release), is that valid? The Rust atomic docs say "The (failed) load ordering can only be [SeqCst](https://doc.rust-lang.org/std/sync/atomic/enum.Ordering.html#variant.SeqCst), [Acquire](https://doc.rust-lang.org/std/sync/atomic/enum.Ordering.html#variant.Acquire) or [Relaxed](https://doc.rust-lang.org/std/sync/atomic/enum.Ordering.html#variant.Relaxed)."

LegNeato: It's hard for me to follow, perhaps better to list out explicit pairs and not use catch-alls?
```rust
match (order, failure_order) {
    (AtomicOrdering::Relaxed, AtomicOrdering::Relaxed)
    | (AtomicOrdering::Acquire, AtomicOrdering::Relaxed | AtomicOrdering::Acquire)
    | (AtomicOrdering::Release, AtomicOrdering::Relaxed)
    | (AtomicOrdering::AcqRel,  AtomicOrdering::Relaxed | AtomicOrdering::Acquire)
    | (AtomicOrdering::SeqCst,  AtomicOrdering::Relaxed | AtomicOrdering::Acquire | AtomicOrdering::SeqCst) => {}
    _ => { self.abort(); self.unreachable(); }
}
```

crates/rustc_codegen_nvvm/src/builder.rs
```diff
@@ -1609,3 +1681,98 @@ impl<'a, 'll, 'tcx> Builder<'a, 'll, 'tcx> {
...
+        // 2. The thread address space(which should be only accessible to 1 thread, anyway?)
+        // So, we do the following:
+        // 1. Check if the pointer is in one of the address spaces atomics support.
+        //  a) if so, we perform an atomic operation
+        // 2. Check if the pointer is in the thread-local address space. If it is, we use non-atomic ops here,
+        // **ASSUMING** only the current thread can access thread-local memory. (FIXME: is this sound?)
+        // 3. If the pointer is not in a supported address space, and is not thread-local, then we bail, and trap.
+
+        // We check if the `dst` pointer is in the `global` address space.
+        let (isspacep_global_ty, isspacep_global_fn) =
```
LegNeato, line 1710: AI says:

The NVPTX Usage Guide says:

“The ‘llvm.nvvm.isspacep.*’ intrinsics determine whether the provided generic pointer references memory which falls within a particular address space.” 
[LLVM Releases](https://releases.llvm.org/20.1.0/docs/NVPTXUsage.html?utm_source=chatgpt.com)

The NVVM IR Specification (CUDA / NVVM 12.1.1) states:

“In NVVM IR, a generic pointer has a pointer type with the address space generic, while a non-generic pointer has a pointer type with a non-generic address space.” 
[NVIDIA Docs](https://docs.nvidia.com/cuda/archive/12.1.1/nvvm-ir-spec/index.html?utm_source=chatgpt.com)

“The pointer must be either a global pointer, a shared pointer, or a generic pointer that points to either the global address space or the shared address space.” (for cmpxchg) 
[NVIDIA Docs](https://docs.nvidia.com/cuda/archive/12.1.1/nvvm-ir-spec/index.html?utm_source=chatgpt.com)

So: the intrinsic expects a generic pointer. If you have a non-generic (address-space-tagged) pointer (global/shared/local/etc.), you need to either have cast it to generic or use the appropriate ptr.gen.to.* / addrspacecast so that the argument’s pointer type is generic. Otherwise you violate the spec.

LegNeato (changes requested):

LegNeato:
@FractalFir sorry, I had to force push while rebasing for some reason

crates/rustc_codegen_nvvm/src/builder.rs
FractalFir: This would be needed, if our pointers weren't all in the default address space anyway. We cast them as soon as we create them, I belive:
https://github.com/Rust-GPU/rust-cuda/blob/ac2674f03208462709f538f07b10e3f8153c93b6/crates/rustc_codegen_nvvm/src/builder.rs#L512

I will add a comment talking about this - it is something to keep in mind if/when we start adding support for more address spaces.

crates/rustc_codegen_nvvm/src/builder.rs
LegNeato: @FractalFir Any thoughts on this? I'd like to land this as I want to build on top of it 🚀

FractalFir:
Should be good to go, once the CI finishes.

LegNeato (approved):
🥂 welcome to rust cuda!

PR #271, opened 2025-09-18
Add a new devcontainer with enabled LLVM assertions

FractalFir:
This is a variant of the `ubuntu24-cuda12` container, which comes with a debug build of LLVM.

 This is very useful for developement, since LLVM assertions will catch some errors ealier. Additionally, the debug symbols just make debugging code easier.


LegNeato (approved):
Thanks! I guess I usually rebuilt it when I needed it but this will help a lot.

PR #272, opened 2025-09-18
Add apt-get retries for docker containers

LegNeato:


FractalFir:
Would setting the retry number in https://linux.die.net/man/5/apt.conf not be a better solution? Seems like that would apply to all calls to apt, which would be a bit more future-prof(if we add new calls to apt, they would also get this behaviour by default). 

Besides that tiny nit, LGTM.

PR #275, opened 2025-09-19
Fix CI when codegen .so is clobbered

LegNeato:
The workspace builds rustc_codegen_nvvm as a normal Cargo member. During every CUDA build script, cuda_builder re-builds that same crate through -Zcodegen-backend=. Cargo produces two identical librustc_codegen_nvvm.so.

Add a feature to differentiate the two cases, cribbing from rust-gpu as it has the same structure but no overlapping .so issue


LegNeato:
Because this is intermittent, I am just going to merge and watch `main`.

PR #276, opened 2025-09-21
Intrinsics will now use fallback bodies if possible.

FractalFir:
*Some* Rust intrinsics provide fall-back implementations. 

That means a codegen *can* overwrite them, but does not have to.

This PR makes it so that we use those fallback impls if they are present, instead of erroring out.


LegNeato (approved):
TIL.

PR #277, opened 2025-09-21
Fixed a bug where self-referential types would cause infinite recursion and stack overflows

FractalFir:
As the title states. Currently, self-referential types would cause stack overflows(because if `Foo` contains a `Foo*`, then that creates an unsatisfiable cycle). 

This PR just uses `u8*` everywhere instead. Typed pointers are a relic of old LLVM, so getting rid of them(even if partially) now can only do us good.


FractalFir:
> These all look fine but the PRs are overlapping?

Yeah, must have screwed something up in git. I have a couple of completely unrelated bug-fixes(couple more on the way), so managing them all has been a challenge. 

Will fix tomorrow.

LegNeato:
@FractalFir is this still relevant?

PR #278, opened 2025-09-22
Emulate all 128 bit ops

LegNeato:
I thought we could use the fallback code from compiler-builtins but I guess not? Most things were broken.

I had AI bang out our implementations, I did not write them.

This is generating a lot of code, perhaps better to generate a fn and then have everything call it?

I added an example and confirmed it works / passes.


LegNeato:
@nnethercote this changes `math_builder_methods`

LegNeato:
@nnethercote I am going to rebase this on your macro changes, they look better.

LegNeato:
I'm still not sure this is right, as I would expect to be using `compiler-builtins` instead of reimplementing the fallback ourselves. But I am a bit out of my depth as to how all this fits together.

LegNeato:
Ok, so what was happening is we were getting called recursively because `compiler-builtins` shift apis call shift intrinsics (?) and we were infinite looping on codegen. Then I only used the `compiler-builtins` division stuff and had AI manually implement the shifts, but for some reason the compiler division stuff was only operating on the lower (?) 64 bits so the CPU vs GPU comparisons were not passing. Which leaves this where it ended up...we manually impl the shifts to prevent recursion, we manually impl division to get around the 64bit only, and the rest we use the `compiler-builtin` exports

PR #294, opened 2025-10-22
Add CUDA IPC functions

andfoy:
This PR exposes the functions `cudaIpcGetMemHandle`, `cudaIpcOpenMemHandle` and `cudaIpcCloseMemHandle`, it also adds the definition for `cudaIpcMemHandle_t`, which are required to share device memory pointers between processes. This can be helpful for cross-language IPC access, e.g., creating a CuPy array in Python and then accessing the values as a `tch::Tensor`.


LegNeato:
Not familiar with this but looks fine, thank you!

PR #298, opened 2025-11-03
Remove size assertion

LegNeato:
This was overzealous according to @FractalFir and I am hitting it as well.


FractalFir:
Should be <= for soundness.

LegNeato:
@FractalFir should we land this?

FractalFir:
Looks good to me, it is a tiny change.

PR #301, opened 2025-11-10
Update README with current docker image path

Snehal-Reddy:
./container/ubuntu24/Dockerfile is no longer an existing Docker directory in the repo.


LegNeato (approved):
Thank you!

PR #316, opened 2025-11-22
Fix rocky linux builds by including deps

LegNeato:
I thought these were only needed for an example,
turns out they are needed during bindgen regardless.


LegNeato:
Again, needs to be on main to be in the registry and then it should pass.

Issue #238, opened 2025-07-26
Port CUDA samples


LegNeato:
We should port these, which will show gaps in our support:

https://github.com/NVIDIA/cuda-samples/

madhav-madhusoodanan:
I'd love to take this up!

madhav-madhusoodanan:
I ran into this error because I ended up allocating more memory than the capacity of my GPU (2GB):

```
relocation R_X86_64_PC32 out of range: -2147576665 is not in [-2147483648, 2147483647]
```

Thought I'd document it here.

PR #341, opened 2025-12-10
Samples: Ported the MatrixMul example of CUDA samples

madhav-madhusoodanan:
# Context
Ported the `matrixMul` sample from `NVIDIA/cuda-samples`.

# Relevant Issue
- https://github.com/Rust-GPU/rust-cuda/issues/238

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,6 +54,8 @@ members = [
 
   "samples/introduction/async_api",
   "samples/introduction/async_api/kernels",
+  "samples/introduction/matmul",
+  "samples/introduction/matmul/kernels",
 
   "tests/compiletests",
   "tests/compiletests/deps-helper",
--- a/samples/introduction/README.md
+++ b/samples/introduction/README.md
@@ -1,8 +1,0 @@
-# Chapter 0: Introduction
-
-## [asyncAPI](https://github.com/Rust-GPU/rust-cuda/samples/introduction/async_api)
-This example demonstrates two key capabilities of CUDA events: measuring GPU execution time and enabling concurrent CPU-GPU operations.
-
-1. Events are recorded at specific points within a CUDA stream to mark the beginning and end of GPU operations.
-2. Because CUDA stream operations execute asynchronously, the CPU remains free to perform other work while the GPU processes tasks (including memory transfers between host and device)
-3. The CPU can query these events to check whether the GPU has finished its work, allowing for coordination between the two processors without blocking the CPU.
--- a/samples/introduction/matmul/kernels/src/lib.rs
+++ b/samples/introduction/matmul/kernels/src/lib.rs
@@ -1,0 +1,80 @@
+use core::mem::MaybeUninit;
+use cuda_std::*;
+
+#[kernel]
+#[allow(improper_ctypes_definitions)]
+#[allow(clippy::needless_range_loop)]
+/// # Safety
+///
+/// This function is unsafe because it dereferences raw pointers.
+pub unsafe fn matrix_mul_cuda(c: *mut f32, a: &[f32], b: &[f32], wa: usize, wb: usize) {
+    let bx: usize = cuda_std::thread::block_idx().x as usize;
+    let by: usize = cuda_std::thread::block_idx().y as usize;
+
+    let tx: usize = cuda_std::thread::thread_idx().x as usize;
+    let ty: usize = cuda_std::thread::thread_idx().y as usize;
+
+    const BLOCK_SIZE: usize = 32;
+    let a_begin = wa * BLOCK_SIZE * by;
+    let a_end = a_begin + wa - 1;
+    let a_step = BLOCK_SIZE;
+
+    let b_begin = BLOCK_SIZE * bx;
+    let b_step = BLOCK_SIZE * wb;
+
+    let mut c_sub: f32 = 0.0;
+    let mut kahan_correction_factor = 0.0f32;
+    let mut bi = b_begin;
+
+    for ai in (a_begin..=a_end).step_by(a_step) {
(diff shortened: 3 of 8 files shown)


madhav-madhusoodanan:
Seems there are other modules that need formatting

Cargo.toml
```diff
@@ -55,6 +55,9 @@ members = [
   "samples/introduction/async_api",
   "samples/introduction/async_api/kernels",
```
nnethercote: No need for a blank line here.

samples/introduction/README.md
```diff
@@ -6,3 +6,11 @@ This example demonstrates two key capabilities of CUDA events: measuring GPU exe
 1. Events are recorded at specific points within a CUDA stream to mark the beginning and end of GPU operations.
 2. Because CUDA stream operations execute asynchronously, the CPU remains free to perform other work while the GPU processes tasks (including memory transfers between host and device)
 3. The CPU can query these events to check whether the GPU has finished its work, allowing for coordination between the two processors without blocking the CPU.
+
+## [matrixMul](https://github.com/Rust-GPU/rust-cuda/samples/introduction/matmul)
+This example demonstrates an example kernel implementation of matrix multiplicaation.
```
nnethercote:
```suggestion
This example demonstrates an example kernel implementation of matrix multiplication.
```

samples/introduction/matmul/kernels/src/lib.rs
```diff
@@ -0,0 +1,78 @@
+use core::mem::MaybeUninit;
+use cuda_std::*;
+
+// SAFETY: This function is unsafe because it dereferences raw pointers.
+#[kernel]
+pub unsafe fn matrix_mul_cuda(C: *mut f32, A: *const f32, B: *const f32, wa: usize, wb: usize) {
```
nnethercote: Probably better to name these `c`, `a`, `b`, given that upper-case names are normally used for statics. You'll need to rename some variables below as well, perhaps `ai` and `bi` would work.

nnethercote: `c` must be a raw pointer because it's modified. But `a` and `b` can be slices because they are const, which would be more idiomatic. Can you change them?

samples/introduction/matmul/src/main.rs
```diff
@@ -0,0 +1,164 @@
...
+
+    let d_a =
+        DeviceBuffer::from_slice(h_a.as_slice()).expect("device array couldn't be initialized!");
+    let d_b =
+        DeviceBuffer::from_slice(h_b.as_slice()).expect("device array couldn't be initialized!");
+    let d_c =
+        DeviceBuffer::from_slice(h_c.as_slice()).expect("device array couldn't be initialized!");
+
+    stream.synchronize().expect("Stream couldn't synchronize!");
+    let blocks = BlockSize::xy(block_size as u32, block_size as u32);
```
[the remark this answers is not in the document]
nnethercote: Also, `blocks` is called `threads` in the original, which I find clearer. (`blocks` makes me think it's the number of blocks, which is what `grid` actually is.)

samples/introduction/README.md
```diff
@@ -6,3 +6,11 @@ This example demonstrates two key capabilities of CUDA events: measuring GPU exe
...
+
+1. The matrices are first created on the host side and then copied to the device.
+2. A shared piece of block-specific memory is created (on the device side), so that summation can be done very quickly
+3. The result is copied back to host, where the accumulated error occur.
+4. Extra: The error that accumulates during the summation process is reduced (in the kernel itself) using [Kahan summation algorithm](https://en.wikipedia.org/wiki/Kahan_summation_algorithm).
```
nnethercote: I wonder if you should eliminate this README and move each per-sample description into a top-level doc comment in the sample's `main.rs`? Reason being that I generally find it useful to have code-specific documentation in the code rather than in a separate file. That way it's easier to find and more likely to be kept up to date.

nnethercote (changes requested):
Thanks for doing this! Generally looking good, though I have a few suggestions.

Also, can you squash the commits together? They conceptually belong together, and there's no point polluting version control history with all these incomplete intermediate revisions.

samples/introduction/matmul/src/main.rs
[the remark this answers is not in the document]
madhav-madhusoodanan: Okay, that makes sense.

samples/introduction/matmul/kernels/src/lib.rs
madhav-madhusoodanan: Won't this cause the below error? I'm running into that in the CI
```
error: `extern` fn uses type `[f32]`, which is not FFI-safe
 --> samples/introduction/matmul/kernels/src/lib.rs:6:47
  |
6 | pub unsafe fn matrix_mul_cuda(c: *mut f32, a: &[f32], b: &[f32], wa: usize, wb: usize) {
  |                                               ^^^^^^ not FFI-safe
  |
  = help: consider using a raw pointer instead
  = note: slices have no C equivalent
```

nnethercote: You can use `#[allow(improper_ctypes_definitions)]` to get around this. https://rust-gpu.github.io/rust-cuda/guide/getting_started.html has more details.

nnethercote:
Sorry for the slow follow-up here. Thanks for the changes you've made. clippy needs fixing, and the commit message should be simplified to something more coherent. Otherwise, looking good!

madhav-madhusoodanan:
Apologies from my side too!
I've just resolved the issues here. Let me know if this is good to go.

nnethercote:
LGTM, thanks.

PR #342, opened 2025-12-11
Fix `path-tracer`.

nnethercote:
The merging of two unintentionally interdependent PRs left `path-tracer` in a broken state.
- #335 updated it from edition 2018 to 2024, while it was disabled (commented out).
- #332 re-enabled it.

This left it marked as using edition 2024 but without the appropriate changes for edition 2024. This commit makes those changes.


LegNeato (approved):
lgtm

PR #345, opened 2025-12-22
examples: Add dot product of two vectors example

karpenkoag:
- An example of a scalar product of two vectors on a GPU has been added.

- In the example, a benchmark of three implementation options is performed: through a cuBLAS call, using native CUDA code, and using the Rust-CUDA variant.


LegNeato:
Thanks for the PR! Seems to be failing on windows, not sure if that is our test setup or your issue.

PR #349, opened 2026-01-14
Update to nightly-2025-08-11.

fee1-dead:


LegNeato:
I think we support f16 and f128 via polyfills? My memory is hazy

LegNeato (approved):
Approving for now.

PR #350, opened 2026-01-14
Allow adjusting loop unroll aggressiveness

niklebedenko:
In C++ CUDA, there's `#pragma unroll`, which allows you to force a loop to be unrolled. Rust does not have such an equivalent, but LLVM will decide whether to unroll your loop based on heuristics. If you set the LLVM option [`-unroll-threshold`](https://llvm.org/doxygen/LoopUnrollPass_8cpp.html#ab5709dc220a64908090b46d1d1f6309b) to a large number, you can make LLVM be more aggressive in its unrolling of loops.

This gave my Rust code **a 10x speed improvement** for some kernels, due to being able to index arrays at compile time, removing the need for local memory, stack frames, and function calls.

N.B. there is the [`unroll` crate](https://docs.rs/unroll/latest/unroll/) but it only supports unrolling loops with integer bounds. The LLVM approach allows unrolling loops over iterators also.

### DIFF
--- a/crates/cuda_std/Cargo.toml
+++ b/crates/cuda_std/Cargo.toml
@@ -8,7 +8,7 @@ repository = "https://github.com/Rust-GPU/rust-cuda"
 readme = "../../README.md"
 
 [dependencies]
-glam = { version = ">=0.22", default-features = false, features = ["libm", "cuda", "bytemuck"] }
+glam = { version = ">=0.27", default-features = false, features = ["libm", "cuda", "bytemuck"] }
 vek = { version = "0.17.1", default-features = false, features = ["libm"] }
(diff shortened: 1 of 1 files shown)


crates/cuda_std/Cargo.toml
```diff
@@ -8,7 +8,7 @@ repository = "https://github.com/Rust-GPU/rust-cuda"
 readme = "../../README.md"
 
 [dependencies]
-glam = { version = ">=0.22", default-features = false, features = ["libm", "cuda", "bytemuck"] }
+glam = { version = ">=0.27", default-features = false, features = ["libm", "cuda", "bytemuck"] }
```
LegNeato: Why was this bumped?

PR #357, opened 2026-02-09
fix(cuda_std): use correct PTX scope suffix in atomic load/store

### LINKED ISSUE #354: Invalid PTX generation for atomic load/store: incorrect scope suffix
First off, thanks for all the effort put into `rust-cuda` and `cuda_std`.

 I've been working with the atomic intrinsics and noticed a small issue in how the PTX assembly is generated for `load` and `store` operations within `crates/cuda_std/src/atomic/intrinsics.rs`.

**The Issue**: The `load_scope!` macro currently uses the Rust-level identifier (`$scope`) instead of the assembly-level identifier (`$scope_asm`) when generating the instruction string.

For example, when generating a relaxed load at the device level:
  - **Current Output:** `ld.relaxed.device.u32 ...` (uses `$scope`: "device")
  - **Expected Output:** `ld.relaxed.gpu.u32 ...` (should use `$scope_asm`: "gpu")

It looks like the other macros in the file (like `atomic_fetch_op_2_reg`) are already correctly using `$scope_asm`.

Snehal-Reddy:
This PR fixes a bug in the `load_scope!` macro in `crates/cuda_std/src/atomic/intrinsics.rs`

Previously, the macro was using the $scope identifier (e.g., device, block) when generating the PTX instruction string. This resulted in invalid assembly instructions like `ld.relaxed.device.u32`.

This change updates the macro to use `$scope_asm` (e.g., gpu, cta), ensuring valid PTX suffixes are generated (e.g., `ld.relaxed.gpu.u32`).

fixes #354

### DIFF
--- a/crates/cuda_std/src/atomic/intrinsics.rs
+++ b/crates/cuda_std/src/atomic/intrinsics.rs
@@ -52,11 +52,11 @@ pub unsafe fn fence_acqrel_system() {
 
 #[allow(unused_macros)]
 macro_rules! load_scope {
-    (volatile, $scope:ident) => {
+    (volatile, $scope_asm:ident) => {
         ""
     };
-    ($ordering:ident, $scope:ident) => {
-        concat!(".", stringify!($scope))
+    ($ordering:ident, $scope_asm:ident) => {
+        concat!(".", stringify!($scope_asm))
     };
 }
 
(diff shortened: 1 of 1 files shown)


LegNeato:
Thanks!

PR #360, opened 2026-02-10
fix(cuda_std): align WarpShuffleMode discriminants with LLVM intrinsics

### LINKED ISSUE #359: Incorrect WarpShuffleMode discriminants mapping to wrong shuffle instructions
### Issue

The `WarpShuffleMode` enum in `crates/cuda_std/src/warp.rs` relies on implicit discriminants. These default values do not match the mode integer arguments expected by the underlying LLVM intrinsic `llvm.nvvm.shfl.sync.i32`. This causes the high-level safe wrappers  (warp_shuffle_up, warp_shuffle_down, warp_shuffle_idx) to invoke the wrong shuffle operations at the PTX level.

  The llvm.nvvm.shfl.sync.i32 intrinsic (and the corresponding PTX `shfl.sync` instruction) expects the mode argument to be:
   * `0`: Idx (shfl.idx)
   * `1`: Up (shfl.up)
   * `2`: Down (shfl.down)
   * `3`: Bfly (shfl.bfly / xor)

However, the previous definition of `WarpShuffleMode` relied on default Rust enum discriminants:

```
 pub enum WarpShuffleMode {
   Up,   // Implicitly 0
   Down, // Implicitly 1
   Idx,  // Implicitly 2
   Xor,  // Implicitly 3
 }
```

 ### Consequences
  This mismatch results in the following incorrect mappings:
   * Calling warp_shuffle_idx passes Mode::Idx (2), which executes `shfl.sync.down`.
   * Calling warp_shuffle_up passes Mode::Up (0), which executes `shfl.sync.idx`.
   * Calling warp_shuffle_down passes Mode::Down (1), which executes `shfl.sync.up`.

  This leads to unpredictable behavior and silent logic errors. For example, broadcasting an index via warp_shuffle_idx results in a
  down-shuffle, often producing 0 or invalid indices that cause Illegal Address errors when used for memory access.

Snehal-Reddy:
The `WarpShuffleMode` enum previously relied on implicit discriminants (Up=0, Down=1, Idx=2, Xor=3), which misaligned with the integer values expected by the `llvm.nvvm.shfl.sync.i32` intrinsic (Idx=0, Up=1, Down=2, Bfly=3).

Fixes #359

### DIFF
--- a/crates/cuda_std/src/warp.rs
+++ b/crates/cuda_std/src/warp.rs
@@ -735,10 +735,10 @@ impl WarpShuffleValue for bf16 {
 #[repr(u32)]
 #[derive(Clone, Copy)]
 pub enum WarpShuffleMode {
-    Up,
-    Down,
-    Idx,
-    Xor,
+    Idx = 0,
+    Up = 1,
+    Down = 2,
+    Xor = 3,
 }
 
 // C-compatible struct to match LLVM IR's {i32, i8} return type


LegNeato:
🍻

PR #382, opened 2026-04-26
fix warp assert on pow of 2

39ali:
## Summary
Rust `!` on integers is `bitwise NOT`, `not logical NOT`, now it should assert correctly on none pow of 2 

- [x] `cargo build` passes
- [ ] `cargo clippy --workspace` passes
- [ ] Tested on: [OS, GPU, CUDA version]

## Notes

Any additional context for reviewers.

### DIFF
--- a/crates/cuda_std/src/warp.rs
+++ b/crates/cuda_std/src/warp.rs
@@ -766,7 +766,7 @@ unsafe fn warp_shuffle_32(
     }
 
     assert!(
-        !(width & (width - 1)) != 0 && width <= 32,
+        width != 0 && (width & (width - 1)) == 0 && width <= 32,
         "width must be a power of 2 and less than or equal to 32"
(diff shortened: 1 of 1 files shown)


crates/cuda_std/src/warp.rs
```diff
@@ -766,7 +766,7 @@ unsafe fn warp_shuffle_32(
     }
 
     assert!(
-        !(width & (width - 1)) != 0 && width <= 32,
+        width != 0 && (width & (width - 1)) == 0 && width <= 32,
```
fee1-dead: Why can't we use https://doc.rust-lang.org/std/primitive.u32.html#method.is_power_of_two here?

39ali: that's true, we should use it, changing it

PR #386, opened 2026-04-29
feat(container): add ubuntu24-cuda13 image with LLVM 19.1.7

brandonros:
part of https://github.com/Rust-GPU/rust-cuda/issues/385

PR #389, opened 2026-05-01
 fix(nvvm): resolve Windows CI build failures

Snehal-Reddy:
## Summary

1. **Broken LLVM Download URL:** The hardcoded download URL incorrectly used the uppercase tag LLVM-7.1.0 (which has no release assets), instead of the correct lowercase tag llvm-7.1.0. Because curl didn't verify the HTTP response code, the script was attempting to unpack a GitHub 404 HTML page as a tarball, resulting in an obscure UnexpectedEof / TarError panic.
2. **Invalid MSVC Compiler Flags:** When the `cc` crate attempts to compile the LLVM C++ shim (`RustWrapper.cpp`) using MSVC (`cl.exe`), it blindly passes all flags returned by `llvm-config --cxxflags`. Since the prebuilt LLVM was cross-compiled, `llvm-config` returns GCC-style warning flags (e.g. `-Wextra`), which causes MSVC to fail with a `D8021 : invalid numeric argument '/Wextra'` error.

## Changes

- Corrected PREBUILT_LLVM_URL_LLVM7 to use the lowercase llvm-7.1.0 tag.
- Added an explicit HTTP response code check to catch download failures immediately. It will now fail with a clear, actionable error message showing the bad URL and response code.
- Added logic to filter out all GCC-style warning flags (`-W*`) when compiling the C++ shim if the target contains "msvc".

## Testing

- [x] cargo build passes
- [x] cargo clippy --workspace passes
- [x] Tested on: Windows / Linux

## Notes

This issue manifested as build failures specifically when running cargo build in a clean Windows environment where USE_PREBUILT_LLVM triggers.

### DIFF
--- a/crates/rustc_codegen_nvvm/build.rs
+++ b/crates/rustc_codegen_nvvm/build.rs
@@ -11,7 +11,7 @@ use tar::Archive;
 use xz::read::XzDecoder;
 
 static PREBUILT_LLVM_URL_LLVM7: &str =
-    "https://github.com/rust-gpu/rustc_codegen_nvvm-llvm/releases/download/LLVM-7.1.0/";
+    "https://github.com/rust-gpu/rustc_codegen_nvvm-llvm/releases/download/llvm-7.1.0/";
 
 fn main() {
     rustc_llvm_build(llvm19_enabled());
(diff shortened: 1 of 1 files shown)


brandonros:
i pulled this into https://github.com/Rust-GPU/rust-cuda/pull/386 if we want

Snehal-Reddy:
Aa great, thanks @brandonros . Let me close this then

PR #391, opened 2026-05-03
fix(cuda_std): pack warp shuffle/match return into i64 for LLVM 19 verifier

brandonros:
if we land https://github.com/Rust-GPU/rust-cuda/pull/386 this becomes smaller, opening it for now

fixes

```
Compiling cuda_std v0.2.2 (https://github.com/brandonros/Rust-CUDA.git?rev=38212ab745b6d257992e37bc0bc2b7b659bebed4#38212ab7)
  error: LLVM module verification failed for cuda_std.ee6756d2761d070-cgu.0: Attribute 'align 4' applied to incompatible type!
           %6 = call align 4 { i32, i8 } @__nvvm_warp_shuffle(i32 %mask, i32 %mode, i32 %value, i32 %b, i32 %5)
         Attribute 'align 4' applied to incompatible type!
         ptr @__nvvm_warp_shuffle
         

  error: could not compile `cuda_std` (lib) due to 1 previous error

  thread 'main' (192541) panicked at cli/build.rs:33:10:
  called `Result::unwrap()` on an `Err` value: BuildFailed
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

```


brandonros:
@LegNeato if we could land this one please, thank you very much

Issue #353, opened 2026-01-23
devtools: mdbook preprocessor, compiler explorer?


j4orz:
I'm working on a [deep learning systems book](https://j4orz.ai/sitp/) (the #teenygrad work group within the GPU MODE discord) and interested in using CUDA Rust over CUDA C++ given my familiarity with the host syntax and semantics of the former. Has anyone on the team thought of making an mdbook preprocessor and/or support for compiler explorer? This is something I'm looking for with my use case (and I also suspect it would benefit people going through the [Rust CUDA Book](https://rust-gpu.github.io/rust-cuda/introduction.html).

LegNeato:
https://github.com/Rust-GPU/rust-cuda/issues/177 for compiler explorer

PR #394, opened 2026-05-06
feat(godbolt): add Compiler Explorer integration for rust-cuda kernels

### LINKED ISSUE #177: Add to godbolt
It would be great to have rust-cuda on https://godbolt.org/.

The repo is here:

https://github.com/compiler-explorer/compiler-explorer

g-ampo:
## Summary

Add Compiler Explorer integration so users can compile rust-cuda kernels to PTX in the browser.

Closes #177, partially addresses #353.

Building on the discussion between @LegNeato and @Forsworns .This takes the "wrapper script that replicates cuda_builder's rustc invocation" route that @LegNeato suggested (to avoiding the need for cargo on the CE side)

## Changes

- `contrib/godbolt/rust-cuda-wrapper.sh` - wrapper that creates a temp Cargo project with the same rustflags as `cuda_builder::invoke_rustc()`, builds to PTX, extracts output from cargo JSON
- `contrib/godbolt/install.sh` - installs pinned nightly, builds `rustc_codegen_nvvm`, lays out a self-contained prefix
- `contrib/godbolt/rust-cuda.defaults.properties` - CE language/compiler config
- `contrib/godbolt/test-kernel.rs` - sample kernels for smoke-testing
- `contrib/godbolt/README.md` - local testing and upstream submission steps

## Testing

- [ ] `cargo check` passes
- [ ] `cargo clippy --workspace` passes
- [x] Tested on: Ubuntu 24.04, RTX 3060 Ti, CUDA 13.0


contrib/godbolt/rust-cuda.defaults.properties
```diff
@@ -0,0 +1,65 @@
...
+groupName=Rust CUDA (rust-cuda)
+
+# The wrapper script that Godbolt executes.  install.sh places it here.
+compiler=/opt/compiler-explorer/rust-cuda/bin/rust-cuda-wrapper.sh
+
+# Language shown in the UI dropdown.
+lang=rust
+
+# Human-readable name in the compiler selector.
+name=rust-cuda (nightly-2026-04-02)
```
LegNeato: The `rustc` version is nightly-2026-04-02, but shouldn't this be rust-cuda `main`? Will it automatically get main or do we have to keep bumping in compiler explorer?

Also we should maybe have llvm7 in the name somewhere so later we can have llvm19+ as an option? Not sure how variants like that are handled.

contrib/godbolt/rust-cuda.defaults.properties
```diff
@@ -0,0 +1,65 @@
...
+
+# Semantic version used for sorting in the compiler list.
+semver=0.3.0
```
LegNeato: What is this version from?

contrib/godbolt/rust-cuda.defaults.properties
```diff
@@ -0,0 +1,65 @@
...
+
+# ---------------------------------------------------------------------------
+# Behaviour
+# ---------------------------------------------------------------------------
+
+# The wrapper always writes PTX to stdout and diagnostics to stderr.
+compilerShouldNotProduceOutput=true
```
LegNeato, line 31: Should this be `false`? The double negative is throwing me a bit.

contrib/godbolt/rust-cuda.defaults.properties
```diff
@@ -0,0 +1,65 @@
...
+
+# No execution support (GPU code cannot run on the CE host).
+supportsExecute=false
+
+# The --version flag is handled by the wrapper.
+versionFlag=--version
+versionRe=rust-cuda-wrapper.*
+
+# Instruction-set reference for hover tooltips.
+instructionSet=ptx
```
LegNeato, line 59: What does this do? https://docs.nvidia.com/cuda/parallel-thread-execution/ is the reference docs FWIW.

contrib/godbolt/rust-cuda.defaults.properties
```diff
@@ -0,0 +1,65 @@
...
+
+# ---------------------------------------------------------------------------
+# Source boilerplate
+# ---------------------------------------------------------------------------
+# The default source shown when a user picks this compiler.
+defaultSource=use cuda_std::prelude::*;\n\n#[kernel]\n#[allow(improper_ctypes_definitions, clippy::missing_safety_doc)]\npub unsafe fn vecadd(a: &[f32], b: &[f32], c: *mut f32) {\n    let idx = thread::index_1d() as usize;\n    if idx < a.len() {\n        let elem = unsafe { &mut *c.add(idx) };\n        *elem = a[idx] + b[idx];\n    }\n}
```
LegNeato, line 65: Will cuda_std be included in compiler explorer? Should this match the test kernel?

contrib/godbolt/rust-cuda-wrapper.sh
```diff
@@ -0,0 +1,241 @@
...
+
+# ---------------------------------------------------------------------------
+# Defaults
+# ---------------------------------------------------------------------------
+RUST_CUDA_ROOT="${RUST_CUDA_ROOT:-/opt/compiler-explorer/rust-cuda}"
+CUDA_PATH="${CUDA_PATH:-/usr/local/cuda}"
+
+EMIT="ptx"
+OPT_LEVEL="3"
+# Default to compute_75 (Turing), matching NvvmArch::default() when llvm19 is off.
```
LegNeato: So this is not using the llvm19 path? Is there a way to switch?

LegNeato (changes requested):
Thanks for working on this! I am not super familiar with compiler explorer, so apologies if my review questions are dumb.

contrib/godbolt/rust-cuda.defaults.properties
g-ampo: CE would re-run install.sh to update which clones from main by default. The name and semver need a manual bump when you tag a new release. About llvm7 yes I will add that to the name :)

contrib/godbolt/rust-cuda.defaults.properties
g-ampo: That's from the [0.3](https://github.com/Rust-GPU/rust-cuda/releases/tag/0.3) tag

contrib/godbolt/rust-cuda.defaults.properties
g-ampo: Hmm the name is confusing ,but `true` is correct here. It tells CE not to look for an output file. The wrapper writes PTX to stdout, and `outputIsAsm=true` tells CE to treat that as the assembly output.

contrib/godbolt/rust-cuda.defaults.properties
g-ampo: It enables hover tooltips for PTX instructions in the CE editor. Got it from that same PTX ISA reference you sent.

contrib/godbolt/rust-cuda.defaults.properties
g-ampo: Yes install.sh copies cuda_std and cuda_std_macros into the prefix. The wrapper creates a temp Cargo project that depends on it. The default source is the simple vecadd from test-kernel.rs. The full test kernel also has a tiled matvec but that felt heavy for a default

contrib/godbolt/rust-cuda-wrapper.sh
g-ampo: Right, llvm7 path only. The llvm19 codegen would need a separate build (different `--features`, Compute100 default, CUDA 12.9+ libnvvm). When that's ready it would be a second CE compiler entry so no switching within the same wrapper.

contrib/godbolt/rust-cuda.defaults.properties
LegNeato: Yeah, that is really old as we haven't done a release in a long time...is there a way we can just have it be `main` or something?

LegNeato:
Thinking about this some more, instead of shell perhaps it would be better to write this tool in rust (a normal binary that can compile with any version of rust). Then we can add tests and such to our CI to make sure we don't inadvertently break the integration. As it is I am worried that we land something and don't know it breaks until we go to update the version in compiler explorer. Thoughts?

g-ampo:
Yes would make sense. The main thing the Rust binary would still do is shell out to cargo build under the hood, so the breakage surface is similar to shell (for the glue parts)... but if it imports cuda_std or hooks into the build logic directly, that gives compile-time  detection when APIs change, which shell can't do :) 

Would be (very) happy to rewrite it in Rust.
Should I do it in this PR or land what's here and follow up?

contrib/godbolt/rust-cuda.defaults.properties
g-ampo: So I checked CE's asSafeVer() in lib/utils.ts - if the semver field contains main or trunk, it sorts highest in the compiler group. So semver=main works and future tagged releases would sort below it. Will update this!

LegNeato:
Up to you!

g-ampo:
@LegNeato  Done in 46e82ec: wrapper is now a stand-alone Rust binary  (plus an integration test that compiles test-kernel.rs and asserts the output looks like PTX ).

contrib/godbolt/rust-cuda-wrapper/src/main.rs
```diff
@@ -0,0 +1,160 @@
...
+
+use anyhow::{Context, Result};
+use clap::Parser;
+use clap::ValueEnum;
+use std::path::PathBuf;
+
+/// Rustflags applied on every build. These mirror the flags `cuda_builder`
+/// passes when invoking rustc directly, so PTX produced through the wrapper
+/// matches what a normal rust-cuda build would emit.
+const STATIC_RUSTFLAGS: [&str; 6] = [
```
LegNeato, line 25: Can the wrapper just use cuda_builder directly so these can't get out of sync? It looks like you can set `DEP_RUSTC_CODEGEN_NVVM_OUT_DIR` or put it in the proper place expected by https://github.com/Rust-GPU/rust-cuda/blob/103a8d56935c4e0885ff7c3d25402319df1a8e00/crates/cuda_builder/src/lib.rs#L494 and it should possibly work?

LegNeato (changes requested):

contrib/godbolt/rust-cuda-wrapper/src/main.rs
g-ampo: There is some drift risk but unfortunatelly a direct cuda_builder dependency doesn't work here:

- cuda_builder -> nvvm -> cust_raw, and cust_raw (links = "cuda") requires a CUDA SDK at build time. The wrapper would no longer build without a full toolchain which is the thing the Rust rewrite was meant to avoid...
- CudaBuilder::build() writes cargo: lines to stdout (it's a build.rs API); the wrapper needs stdout to be PTX only for CE. The flag logic in invoke_rustc is private anyway. DEP_RUSTC_CODEGEN_NVVM_OUT_DIR only changes where the backend .so is found, not either blocker.

here the real fix would be to extract the flag-building from invoke_rustc into a small dep-free crate that both cuda_builder and the wrapper call so there would be one source of truth. If thats fine with you I would just do this here in this PR.

LegNeato: Sure!

PR #400, opened 2026-07-20
fix(atomic): use system memory scope for SystemAtomic types

### LINKED ISSUE #399: Data Races & Flawed Synchronization due to Invalid Atomic Memory Scope in `SystemAtomicF32` and `SystemAtomicF64
`SystemAtomicF32` and `SystemAtomicF64` types are mistakenly configured with `device` scope instead of `system` scope in their macro instantiations.

This presents the following issues:
1. **Flawed Synchronization**: Code that uses `SystemAtomicF32` or `SystemAtomicF64` intends to synchronize memory accesses across the entire system (between the GPU and host CPU or across multiple GPUs over PCIe/NVLink).
2. **Data Races**: Because the emitted PTX operations use the `.gpu` scope (`atomic.global.gpu...`) rather than the proper `.sys` scope (`atomic.global.sys...`), CPU host accesses and multi-GPU accesses to system memory will not observe proper cache coherency. This leads to data races and corrupted shared memory states without emitting any compilation errors.

## Reproduction Case
Using `SystemAtomicF32` or `SystemAtomicF64` and calling atomic operations (like `fetch_add`, `load`, or `store`) will incorrectly generate device-scoped PTX:

```rust
use cuda_std::atomic::SystemAtomicF32;
use core::sync::atomic::Ordering;

#[cuda_std::kernel]
pub unsafe fn system_atomic_kernel(val: &SystemAtomicF32) {
    // PTX generates: atomic.global.gpu.add.f32 ...
    // Expected: atomic.global.sys.add.f32 ...
    val.fetch_add(1.0, Ordering::Relaxed);
}
```

## Issue Details
In `crates/cuda_std/src/atomic.rs`, the type instantiations for system-level float atomics incorrectly pass the `$scope` argument as `device` instead of `system` to the macro:

```rust
// Current instantiations in atomic.rs
atomic_float!(f32, AtomicF32, 4, device, 32);
atomic_float!(f64, AtomicF64, 8, device, 64);
atomic_float!(f32, BlockAtomicF32, 4, block, 32, unsafe);
atomic_float!(f64, BlockAtomicF64, 8, block, 64, unsafe);
atomic_float!(f32, SystemAtomicF32, 4, device, 32); // <--- BUG: device instead of system
atomic_float!(f64, SystemAtomicF64, 8, device, 64); // <--- BUG: device instead of system
```

Snehal-Reddy:
## Summary
  
`SystemAtomicF32` and `SystemAtomicF64` previously used the `device` domain scope instead of the intended `system` scope in their macro instantiations. This mismatch caused `.gpu` scoped atomic PTX instructions to be emitted (`atomic.  global.gpu...`) rather than `.sys` scoped instructions. 
  
This bug meant that any system-level synchronization (between the host CPU and GPU, or across multiple GPUs over PCIe/NVLink) would fail to observe proper cache coherency, silently causing data races and corrupted shared memory states. This PR corrects the macro arguments to use the correct `system` scope for system atomics.

Closes #399

## Changes

- Changed the `$scope` argument for `SystemAtomicF32` and `SystemAtomicF64` from `device` to `system` in `crates/cuda_std/src/atomic.rs`.

## Testing

- [x] `cargo build` passes
- [x] `cargo clippy --workspace` passes
- [x] Tested on: Ubuntu 24.04.1 LTS, NVIDIA GeForce RTX 5060 Ti, CUDA 12.8.93

### DIFF
--- a/crates/cuda_std/src/atomic.rs
+++ b/crates/cuda_std/src/atomic.rs
@@ -252,5 +252,5 @@ atomic_float!(f32, AtomicF32, 4, device, 32);
 atomic_float!(f64, AtomicF64, 8, device, 64);
 atomic_float!(f32, BlockAtomicF32, 4, block, 32, unsafe);
 atomic_float!(f64, BlockAtomicF64, 8, block, 64, unsafe);
-atomic_float!(f32, SystemAtomicF32, 4, device, 32);
-atomic_float!(f64, SystemAtomicF64, 8, device, 64);
+atomic_float!(f32, SystemAtomicF32, 4, system, 32);
+atomic_float!(f64, SystemAtomicF64, 8, system, 64);


Snehal-Reddy:
https://github.com/Rust-GPU/rust-cuda/pull/386 should now take care of windows CI failures.

PR #408, opened 2026-09-13
Upgrade modern backend to LLVM 21.1.8 and CUDA 13.3

brandonros:
Replace the modern LLVM 19 backend feature with LLVM 21 while retaining LLVM 7 for the legacy path. Update feature forwarding, LLVM tool discovery, the vector-add example, and documentation together. The modern Nix shell uses LLVM 21.1.8 and CUDA 13.3; CUDA 13 Docker definitions use CUDA 13.3.1 bases. Update the Windows CUDA installer action to support CUDA 13.3.1.

LLVM 21 compatibility fixes included here:

- `rustc_llvm_wrapper`: guard the `llvm/Transforms/Instrumentation.h` include to LLVM < 19 (the header is gone in LLVM 21; only the legacy pass registry uses it) and update the wrapper to the LLVM 21 APIs. `tests/llvm-wrapper/` adds a small C++ check of the attribute/capture semantics the wrapper relies on.
- `nvvm`: restore kernel annotations after the LLVM 21 bitcode upgrade.
- `cust`: handle CUDA 13.3's new `CU_GRAPH_NODE_TYPE_RESERVED_16` graph node type, which made an existing match non-exhaustive.

Extend the existing Linux CI matrix with LLVM 21 vector-add host/PTX compilation on x86_64 and aarch64, using CUDA 13.3.1 and LLVM 21.1.8 prebuilts. The six existing LLVM 7 configurations retain their workspace checks. There is no separate experiment workflow.

## LLVM 21 prebuilts

The backend's default LLVM 21 download URL is the upstream `Rust-GPU/rustc_codegen_nvvm-llvm` `llvm-21.1.8` release, which does not exist yet, so default downloads fail until it is published. That needs https://github.com/Rust-GPU/rustc_codegen_nvvm-llvm/pull/4 merged and a release with `linux-x86_64.tar.xz`, `linux-aarch64.tar.xz`, and `windows-x86_64.tar.xz`.

Until then, the archive directory is overridable through `PREBUILT_LLVM_URL` (environment variable for the backend, repository Actions variable for CI; documented in the guide). The CI LLVM 21 jobs currently default to the same three archives published at https://github.com/brandonros/rustc_codegen_nvvm-llvm/releases/tag/llvm-21.1.8. Before merge, that default and the guide example should point at the upstream release.

## Status

All LLVM 7 Linux jobs, both Windows jobs, and all container image builds pass, including the new `Ubuntu-24.04/CUDA-13.3.1/LLVM-21.1.8` image.

[...]

PR #413, opened 2026-09-22
Point LLVM 21 prebuilt URLs at the upstream release

brandonros:
Follow-up to #408. The upstream `Rust-GPU/rustc_codegen_nvvm-llvm` `llvm-21.1.8` release is now published with all three archives, so the Linux CI fallback `PREBUILT_LLVM_URL` and the guide example now point at it instead of the `brandonros` fork. `build.rs` already defaulted to upstream.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


brandonros:
@LegNeato oops :) we probably want this

brandonros:
Added 44ff9e0 to fix the `Create manifest for Ubuntu-24.04/CUDA-13.3.1` failure on main ([run](https://github.com/Rust-GPU/rust-cuda/actions/runs/35671976862/job/106581970153)):

```
ERROR: ghcr.io/rust-gpu/rust-cuda-ubuntu24-cuda13@sha256:5870e0e8...: not found
```

**Cause:** each build job uploads its digest as `digests-<image>-<arch>`, and the merge job downloads `digests-<image>-*`. For `rust-cuda-ubuntu24-cuda13`, that pattern also matches the `rust-cuda-ubuntu24-cuda13-llvm21` artifacts. The download step fetched 4 artifacts instead of 2:

```
- digests-rust-cuda-ubuntu24-cuda13-llvm21-amd64
- digests-rust-cuda-ubuntu24-cuda13-llvm21-arm64
- digests-rust-cuda-ubuntu24-cuda13-arm64
- digests-rust-cuda-ubuntu24-cuda13-amd64
```

`imagetools create` then looked for the llvm21 digests in the `rust-cuda-ubuntu24-cuda13` repo, where they don't exist. The other manifest jobs passed because no other image name starts with theirs.

**Fix:** put the arch first (`digests-<arch>-<image>`) and download `digests-*-<image>`. The image name is now the end of the pattern, so `-llvm21` artifacts no longer match.

These steps only run on non-PR events, so this PR's CI won't test the change. The first real test is the next push to main.

